PR #3, opened 2021-12-30
Setup framework structure for Random Variables

aadiraju:


Opeyem1a (approved):
I think we might be able to improve small pieces of this, but it's not critical right now

PR #9, opened 2022-01-03
Merge frameworks

Opeyem1a:
+ re-implement random testing functionality
+ add chi square testing for uniform distributions
+ regex sentence structure and usage fleshed out
+ utils and wrappers for regex clauses added


src/global/BaseRandomTest.java
```diff
@@ -0,0 +1,105 @@
...
+        if (alreadyFailed) fail();  // TODO: No message (?)
+
+        String output = getOutput();
+        // TODO: not always println though (in general how do we handle multi line outputs?) - arrays for each line in regex sentence?
+        Matcher matcher = getMatches(output, processRegexForPrintlnOutput(combineRegex(getRegexSentence())));
+        if (matcher.find()) {
+            for (int i = 0; i < getRegexSentence().length; i++) {
+                Regexable regexClause = getRegexSentence()[i];
+                if (!(regexClause instanceof RandomVariable)) continue;
+                RandomVariable<?> randomVariable = (RandomVariable<?>) regexClause;
```
aadiraju, line 63: I've seen the exact same conditional being used multiple times, so would it be better to make a helper function for this?

src/global/utils/RandomUtil.java
```diff
@@ -0,0 +1,113 @@
...
+
+        assertWithinRange(value, lower, upper, "One or more of your randomly generated numbers fall outside of the required range.");
+
+        while (binLower < upper) {
+            double binUpper = binLower + gap;
+            if (binLower <= value && value < binUpper)
+                return count;
+            count++;
+            binLower = binUpper;
+        }
```
aadiraju, line 91: This works but for efficiency's sake, I think since this will traverse every single bin one at a time from the start it might hamper the runtime sometimes. I think if we leave the bins as a constant then technically it is O(1) but in case we change it, I want to see if there is a "mathy" way to make this quicker.

aadiraju (approved):

aadiraju:
Looks good. Just some lingering comments that aren't too serious

src/global/BaseRandomTest.java
Opeyem1a: I feel like it'd be a waste to write a function that just does a 1 line operation yknow? Unless we think it might be potentially more complex in the future

src/global/utils/RandomUtil.java
Opeyem1a: Made #11 to resolve this 👍🏾

PR #16, opened 2022-01-18
Added Get Item by Name

### LINKED ISSUE #7: get casted value by index (or name)
The name part is a wishlist thing where every RandomVariable can have a name, and they can get them by name as well

aadiraju:
Change Clause to an abstract class, and add name parameter. Added getItemByName with O(n) implementation. Might need to be improved later, but this works as intended for now.

Closes #7

### DIFF
--- a/src/global/BaseTest.java
+++ b/src/global/BaseTest.java
@@ -39,11 +39,22 @@ public abstract class BaseTest {
         try {
             if (matcher.find()) return matcher.group(index);
             else fail("Your code's output did not follow the correct structure/syntax");
-        } catch (ArrayIndexOutOfBoundsException e) {
-            fail("The specified group doesn't exist");
+        } catch (IndexOutOfBoundsException e) {
+            fail("The specified group (" + (index) + ") doesn't exist");
         }
         return "";  // TODO: logically how does this behave?
     }
+
+    public String getItemByName(String name) {
+        Clause[] regSen = getRegexSentence();
+        int index = -1;
+        for (int i = 0; i < regSen.length; i++) {
+            if (regSen[i].getName().equals(name))
+                index = i + 1;
+        }
+        return getItemAtIndex(index);
+    }
+
 
     // Utilities
     // TODO: if possible, make this getTestedClass() and then use reflections to call .main() inside executeMain()
--- a/src/global/BaseTest.java
+++ b/src/global/BaseTest.java
@@ -40,19 +40,19 @@ public abstract class BaseTest {
             if (matcher.find()) return matcher.group(index);
             else fail("Your code's output did not follow the correct structure/syntax");
         } catch (IndexOutOfBoundsException e) {
-            fail("The specified group (" + (index) + ") doesn't exist");
+            fail("The specified group doesn't exist");
         }
         return "";  // TODO: logically how does this behave?
     }
 
     public String getItemByName(String name) {
         Clause[] regSen = getRegexSentence();
-        int index = -1;
         for (int i = 0; i < regSen.length; i++) {
             if (regSen[i].getName().equals(name))
-                index = i + 1;
+                return getItemAtIndex(i + 1);
         }
-        return getItemAtIndex(index);
+        fail("The specified group doesn't exist");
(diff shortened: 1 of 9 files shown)


src/global/BaseTest.java
```diff
@@ -39,12 +39,23 @@ public String getItemAtIndex(int index) {
...
+        } catch (IndexOutOfBoundsException e) {
+            fail("The specified group (" + (index) + ") doesn't exist");
         }
         return "";  // TODO: logically how does this behave?
     }
 
+    public String getItemByName(String name) {
+        Clause[] regSen = getRegexSentence();
+        int index = -1;
+        for (int i = 0; i < regSen.length; i++) {
```
Opeyem1a, line 51: No need for the `index` variable I think. Just return immediately on line 53 with `i + 1` as the argument.

For error handling just manually throw the error on line 55

src/global/variables/Clause.java
```diff
@@ -1,5 +1,17 @@
...
-public interface Clause extends Regexable {
+public abstract class Clause implements Regexable {
+    String name;
 
+    public Clause(String name) {
+        this.name = name;
+    }
+
+    public Clause() {
+        this.name = "";
```
Opeyem1a, line 11: The name should be null when none is set

src/global/variables/Clause.java
```diff
@@ -1,5 +1,17 @@
 package global.variables;
 
-public interface Clause extends Regexable {
+public abstract class Clause implements Regexable {
```
Opeyem1a, line 3: Did we need to do this?

src/global/variables/RandomDouble.java
```diff
@@ -5,27 +5,57 @@
...
+    }
+
+    public RandomDouble(double lower, double upper, int precision, String name) {
+        super(name);
         this.lower = lower;
         this.upper = upper;
         this.precision = precision;
     }
 
     public RandomDouble(int lower, int upper, int precision) {
```
Opeyem1a, line 50: The `int` ones could just call `RandomDouble(all of their args)` and then it'll overload and call the double constructors by default, right?

src/global/BaseTest.java
```diff
@@ -39,12 +39,23 @@ public String getItemAtIndex(int index) {
...
             else fail("Your code's output did not follow the correct structure/syntax");
-        } catch (ArrayIndexOutOfBoundsException e) {
-            fail("The specified group doesn't exist");
+        } catch (IndexOutOfBoundsException e) {
+            fail("The specified group (" + (index) + ") doesn't exist");
         }
         return "";  // TODO: logically how does this behave?
     }
 
+    public String getItemByName(String name) {
```
Opeyem1a, line 48: Each Clause already has a built-in function to cast the string from the group thing into the correct type it needs to be. I think it'd be super beneficial to carry that through so this method returns the correct type.

Idk how tough this is though because Java is baby

Opeyem1a (changes requested):

src/global/variables/Clause.java
aadiraju: We needed the abstract class because we need a property that is common to all `Clause` objects, otherwise, we'd have to go add a `name` variable to every subclass of `Clause` to achieve the same effect.

src/global/variables/RandomDouble.java
aadiraju: It won't call the double unless atleast one of the values in casted explicitly

src/global/BaseTest.java
aadiraju: opened #17

src/global/variables/Clause.java
Opeyem1a: Oh I see, aye aye

src/global/variables/RandomDouble.java
Opeyem1a: So like
```java
public RandomDouble(int lower, int upper, int precision) {
     RandomDouble(lower, upper, precision, name);
}```
doesn't work here?

Opeyem1a (approved):

PR #20, opened 2022-01-28
Adding whitespace

Opeyem1a:


src/global/BaseTest.java
```diff
@@ -34,6 +34,7 @@ public void setRegexSentence(Clause[] regexSentence) {
         return this.regexSentence;
     }
 
+
     public String getItemAtIndex(int index) {
         Matcher matcher = getMatches(getOutput(), processRegexForPrintlnOutput(combineRegex(getRegexSentence())));
         try {
```
Opeyem1a, line 40: Interesting, not sure if this is the move

Opeyem1a: Actually I think it is the move

PR #21, opened 2022-02-08
Restructure clauses validation

Opeyem1a:
+ Renaming
+ Moved validation code into each clause 
+ Added validateRandom() logic to each RandomClause
+ Restructure code
+ Better generalize RandomUtil methods
+ Formatting
+ Minor changes


src/global/tools/CustomAssertions.java
```diff
@@ -0,0 +1,20 @@
+package global.tools;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class CustomAssertions {
+    // Underscore indicates a custom assertion is being used
+    public static void _assertTrue(boolean condition, String userMessage, String devMessage) {
+        assertTrue(condition, userMessage);
+        // Or some way to make the dev see this message without it being returned as an assertion failure message
+        System.out.println(devMessage);
```
aadiraju, line 10: We might have to look into how we do this

aadiraju (approved):
This structure is much cleaner. It looks good!

src/global/tools/CustomAssertions.java
Opeyem1a: Tbh I guess this is already handled by #31

PR #26, opened 2022-02-08
Integer Literal + Minor Fixes

### LINKED ISSUE #27: Integer literal

Opeyem1a:
+ rename expected output to testSentence()
+ added integer literal
+ fix getByName()

Closes #27

### DIFF
--- a/src/global/variables/IntegerLiteral.java
+++ b/src/global/variables/IntegerLiteral.java
@@ -1,0 +1,24 @@
+package global.variables;
+
+public class IntegerLiteral extends Clause {
+    int value;
+
+    public IntegerLiteral(int value) {
+        super();
+        this.value = value;
+    }
+
+    public IntegerLiteral(int value, String name) {
+        super(name);
+        this.value = value;
+    }
+
+    public int getValue() {
+        return value;
+    }
+
+    @Override
+    public String getRegex() {
+        return "(" + getValue() + ")";
+    }
+}
--- a/src/temp/ClauseExample.java
+++ b/src/temp/ClauseExample.java
@@ -1,0 +1,33 @@
+package temp;
+
+import global.BaseTest;
+import global.variables.*;
+import global.variables.wrappers.Optional;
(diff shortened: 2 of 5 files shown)


src/temp/ClauseExample.java
```diff
@@ -0,0 +1,33 @@
...
+        Example.main(new String[]{});
+    }
+
+    @RepeatedTest(50)
+    public void retrieveValueTest() {
+//        int randomIntegerValue = Integer.parseInt(getItemAtIndex(3));
+        if (getItemAtIndex(3) != null) {
+            int randomIntegerValue = Integer.parseInt(getItemAtIndex(3));
+            System.out.println(randomIntegerValue);
+            assertEquals(11, randomIntegerValue);
```
aadiraju, line 30: Wait what does this test do? Does it assert that the randomIntegerValue should be 11?  but won't it be a random integer from 0 to 9? So this test would fail always?

src/global/variables/IntegerLiteral.java
```diff
@@ -0,0 +1,24 @@
+package global.variables;
+
+public class IntegerLiteral extends Clause {
```
aadiraju, line 3: Do you think it makes sense to have a float/double literal as well? Not sure if there is a use case, but I'm wondering if we'd need it

aadiraju (approved):
This looks good. Just some minor concerns I've outlined in the comments

src/temp/ClauseExample.java
Opeyem1a: Yeah this is just test code so I wanted it to fail so I could see the value it retrieved

aadiraju: Okay perfect

src/global/variables/IntegerLiteral.java
Opeyem1a: We were planning to add them already, I just wanted to get that ball rolling

PR #28, opened 2022-02-08
Add name validation to sentence so there can't be any duplicate names

### LINKED ISSUE #24: Add validation so names can't be duplicated in the same test sentence

aadiraju:
closes #24


src/global/BaseTest.java
```diff
@@ -27,6 +29,16 @@
     public abstract Clause[] testSentence();
 
     public void setRegexSentence(Clause[] regexSentence) {
+        //By using a set we can ensure that there are no duplicates in the regexSentence
+        Set<String> namesSet = new HashSet<>();
+        for (Clause clause : regexSentence) {
+            if (clause.getName() != null) {
+                if (namesSet.contains(clause.getName())) {
+                    fail("The name " + clause.getName() + " is already in use. make sure all names are unique");
```
Opeyem1a, line 37: I think is okay for now, but we should make distinctions between messages for developers to fix a test and messages for the student to see. Student might be confused if they ever see a message like this (even if it should never happen in theory)

Opeyem1a (changes requested):

src/global/BaseTest.java
Opeyem1a: Probably make a `_fail(String userMessage, String devMessage)` in `CustomAssertions` so we can say to the user like "There is an issue with the automated testing of this question" or something

aadiraju: Done

PR #29, opened 2022-02-08
Add manual ordering so our tests work first

### LINKED ISSUE #22: Use @Order(n) to ensure our tests run first

aadiraju:
closes #22


src/global/BaseRandomTest.java
```diff
@@ -45,6 +45,7 @@ public void addValueToMapList(RandomVariable<?> randomVariable, int mapKey, Obje
 
     // TODO: choose number
     @RepeatedTest(500)
+    @Order(1)
```
Opeyem1a, line 48: What's the behaviour here? Since the `BaseTest` test also has `@Order(1)`

Opeyem1a (changes requested):

PR #31, opened 2022-02-11
Setup logging

Opeyem1a:
Added logging so we can print messages to console while testing. This lets us output extra information without needing it to be in the asserting `message` parameter


aadiraju (approved):
This is a smart way of handling developer messages as well.

PR #36, opened 2022-02-18
Pre defined classes programs with changing behavior hard q2

Hedgemon4:


src/pre_defined_classes/programs_with_changing_behaviour/hard/q2/Question2Test.java
```diff
@@ -0,0 +1,25 @@
...
+                new RandomInteger(1, 10),
+                new StringLiteral(" into "),
+                new IntegerLiteral(1000000, 100000000)
+        };
+    }
+
+    public void runMain() {
+        Question2.main(new String[0]);
+    }
+}
```
SamamthaH, line 25: I think this distractor:
int encrypt = crypt / gen.nextInt( 10 );
could still cause this test to pass, maybe still include a test that checks the math?

src/pre_defined_classes/programs_with_changing_behaviour/hard/q2/Question2Test.java
```diff
@@ -0,0 +1,25 @@
+package pre_defined_classes.programs_with_changing_behaviour.hard.q2;
+
+import global.BaseRandomTest;
+import global.BaseTest;
+import global.variables.Clause;
+import global.variables.clauses.IntegerLiteral;
+import global.variables.clauses.RandomInteger;
+import global.variables.clauses.StringLiteral;
+
+public class Question2Test extends BaseRandomTest {
```
Opeyem1a, line 10: I feel like we should just name this class `MainTest` always since this is what it needs to be called when we paste them into the website

src/pre_defined_classes/programs_with_changing_behaviour/hard/q2/Question2Test.java
Opeyem1a: `gen.nextInt( 10 );` could return `0` so this solution would fail if we ran it enough times (which BaseRandomTest does) just because dividing by 0 is an error.

src/pre_defined_classes/programs_with_changing_behaviour/hard/q2/Question2.java
```diff
@@ -0,0 +1,19 @@
+package pre_defined_classes.programs_with_changing_behaviour.hard.q2;
+import java.util.Random;
+
+/*
+Make a program that will generate a random <strong>8 digit</strong> number and encrypt it by dividing it by a second
+random <strong>1-digit</strong> number. At the end of the program, print out these numbers.
+Sample Output: 68056441 has been encrypted using 5 into 13611288
+ */
+
+public class Question2 {
```
Opeyem1a, line 10: We should name these classes what they are called in the question (unless this is the actual name?)

Opeyem1a (changes requested):

src/pre_defined_classes/programs_with_changing_behaviour/hard/q2/Question2Test.java
SamamthaH: Oh, so because there's one random in it, it runs the whole test a bunch of times, not just that one line?

Opeyem1a: Yeee anything that `extends BaseRandomTest` gets run somewhere around 500 times

src/pre_defined_classes/programs_with_changing_behaviour/hard/q2/Question2Test.java
Hedgemon4: Should I name it MainTest instead of whatever the class is called in the question?

src/pre_defined_classes/programs_with_changing_behaviour/hard/q2/Question2Test.java
Hedgemon4: Should I warp it in a try-catch then?

Hedgemon4: I inserted a try-catch block, so that should deal with that issue.

src/pre_defined_classes/programs_with_changing_behaviour/hard/q2/Question2.java
Hedgemon4: I changed the class names to what they are on the document

src/pre_defined_classes/programs_with_changing_behaviour/hard/q2/DeCrypticTest.java
```diff
@@ -0,0 +1,31 @@
...
+                new StringLiteral(" into "),
+                new IntegerLiteral(1000000, 100000000)
+        };
+    }
+
+    public void runMain() {
+        try {
+            DeCryptic.main(new String[0]);
+        }catch(ArithmeticException e){
+            fail("Tried to divide by 0. Make sure your key is larger than 0");
```
Opeyem1a, line 27: Do we need this? The website catches errors and fails the test for them I believe (@aadiraju could you confirm?)

src/pre_defined_classes/programs_with_changing_behaviour/hard/q2/Question2Test.java
Opeyem1a: `MainTest` is the move

src/pre_defined_classes/programs_with_changing_behaviour/hard/q2/MainTest.java
```diff
@@ -0,0 +1,27 @@
...
+                new RandomInteger(1, 10),
+                new StringLiteral(" into "),
+                new IntegerLiteral(1000000, 100000000)
+        };
+    }
+
+    public void runMain() {
+        DeCryptic.main(new String[0]);
+    }
+
```
SamamthaH, line 26: You can check the math on this code if you assign names to the integers

Opeyem1a (dismissed):
First merge maybe 👀
I'll let @aadiraju look it over too though

src/pre_defined_classes/programs_with_changing_behaviour/hard/q2/MainTest.java
```diff
@@ -0,0 +1,36 @@
...
+import global.variables.clauses.IntegerLiteral;
+import global.variables.clauses.RandomInteger;
+import global.variables.clauses.StringLiteral;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class MainTest extends BaseRandomTest {
+    //Parsons with Distractors
```
aadiraju, line 14: Maybe add a space before the comment just for perpetuity and consistency

Hedgemon4: Ok I did this

aadiraju (approved):
Looks good!

PR #37, opened 2022-02-18
Extend IntegerLiteral to include ranges

### LINKED ISSUE #35: Literal extensions
+ literals need to handle a set of valid values
+ or ranges where applicable

aadiraju:
closes #35

### DIFF
--- a/src/global/tools/InvalidClauseException.java
+++ b/src/global/tools/InvalidClauseException.java
@@ -1,0 +1,8 @@
+package global.tools;
+
+public class InvalidClauseException extends Exception {
+
+    public InvalidClauseException(String message) {
+        super(message);
+    }
+}
--- a/src/global/tools/InvalidClauseException.java
+++ b/src/global/tools/InvalidClauseException.java
@@ -1,7 +1,6 @@
 package global.tools;
 
 public class InvalidClauseException extends Exception {
-
     public InvalidClauseException(String message) {
         super(message);
     }
--- a/src/global/variables/clauses/IntegerLiteral.java
+++ b/src/global/variables/clauses/IntegerLiteral.java
@@ -4,34 +4,61 @@ import global.tools.InvalidClauseException;
 import global.utils.RegexUtil;
 import global.variables.Clause;
 
+import java.util.HashMap;
+import java.util.Map;
+
 public class IntegerLiteral extends Clause {
-    int value = 0;
+    private Map<String, Boolean> MODES;
+    int value;
     int upper;
     int lower;
 
+    public void initalizeModeMap() {
+        this.MODES = new HashMap<>();
+    }
+
+    public IntegerLiteral() {
+        super();
+        initalizeModeMap();
+        this.MODES.put("RANGE", true);
+        this.lower = Integer.MIN_VALUE;
+        this.upper = Integer.MAX_VALUE;
+    }
+
+    public IntegerLiteral(String name) {
+        super(name);
+        initalizeModeMap();
+        this.MODES.put("RANGE", true);
+        this.lower = Integer.MIN_VALUE;
+        this.upper = Integer.MAX_VALUE;
+    }
+
     public IntegerLiteral(int value) {
         super();
+        initalizeModeMap();
+        this.MODES.put("VALUE", true);
         this.value = value;
-        this.upper = value;
-        this.lower = value;
-
     }
 
     public IntegerLiteral(int value, String name) {
         super(name);
+        initalizeModeMap();
+        this.MODES.put("VALUE", true);
         this.value = value;
-        this.upper = value;
-        this.lower = value;
     }
 
     public IntegerLiteral(int lower, int upper) {
         super();
+        initalizeModeMap();
+        this.MODES.put("RANGE", true);
         this.lower = lower;
         this.upper = upper;
     }
 
     public IntegerLiteral(int lower, int upper, String name) {
         super(name);
+        initalizeModeMap();
+        this.MODES.put("RANGE", true);
         this.lower = lower;
         this.upper = upper;
     }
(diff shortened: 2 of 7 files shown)


src/global/variables/clauses/IntegerLiteral.java
```diff
@@ -1,26 +1,67 @@
...
     public IntegerLiteral(int value, String name) {
         super(name);
         this.value = value;
+        this.upper = value;
+        this.lower = value;
+    }
+
+    public IntegerLiteral(int lower, int upper) {
+        super();
+        this.value = 0;
```
Opeyem1a, line 28: Why set this value at all here?

src/global/variables/clauses/IntegerLiteral.java
```diff
@@ -1,26 +1,67 @@
...
 
     public int getValue() {
         return value;
     }
 
+    public String getValueOrRegexRange() {
+        if (getLower() == getUpper()) {
+            return String.valueOf(getValue());
+        } else if (getLower() > getUpper()) {
+            //TODO: I'm not sure if having a default like this is a good idea, or if I should put it into the constructor
```
Opeyem1a, line 56: If lower is above upper, then it's configured incorrectly and shouldn't be allowed to reach this step

Opeyem1a: So clauses have a `.validate()` method in the abstract class that default just returns `true`. This can be overridden in this class. For context, the intent of `.validate()` is validating that the output retrieved for this clause is okay, so you can override it to be like "if a lower and upper are defined, return true if the output-retrieved-value is within that, if not then just default return true".

So this ^ handles checking that the number is in the range. We should also have a new constructor here that takes an array of integers so they can set a discrete set of potential values.

We should also make a new method in `Clause.java` called `.validateParams()` that checks if parameters are okay. It should return true by default but in cases like this, we override it to check that upper > lower and such. We then loop through all clauses and ensure that `.validateParams()` is okay (do this @beforeAll tests and raise a custom `ClauseException` if a clause is invalidly configured.

Opeyem1a (changes requested):

src/global/variables/clauses/IntegerLiteral.java
aadiraju: Fair point. I can probably just have a default

Opeyem1a: I think just leave it null though, if the literal has a range then it doesn't need to store a value for any reason, right?

src/global/tools/InvalidClauseException.java
```diff
@@ -0,0 +1,8 @@
+package global.tools;
+
+public class InvalidClauseException extends Exception {
+
```
Opeyem1a, line 4: idk why but this space bothers me

src/global/variables/clauses/IntegerLiteral.java
```diff
@@ -1,26 +1,81 @@
 package global.variables.clauses;
 
+import global.tools.InvalidClauseException;
+import global.utils.RegexUtil;
 import global.variables.Clause;
 
 public class IntegerLiteral extends Clause {
-    int value;
+    int value = 0;
```
Opeyem1a, line 8: Can't we just leave this null? I think everything should be null unless set explicitly in a constructor

src/global/variables/clauses/IntegerLiteral.java
```diff
@@ -1,26 +1,81 @@
...
+        if (getLower() != getUpper()) {
+            if (getLower() > getUpper()) {
+                throw new InvalidClauseException("Lower bound cannot be greater than upper bound");
+            }
+        }
+    }
+
+    public String getValueOrRegexRange() {
+        if (getLower() == getUpper()) {
+            return String.valueOf(getValue());
```
Opeyem1a, line 71: I think the logic here should just be "if upper and lower are defined, use them for regex, if they're not and there is a defined value, use that instead"

src/global/variables/clauses/IntegerLiteral.java
```diff
@@ -1,26 +1,81 @@
...
 
     public int getValue() {
         return value;
     }
 
+    @Override
+    public boolean validate(String matchGroupString) {
+        if (getLower() != getUpper()) {
+            int matchGroupInt = Integer.parseInt(matchGroupString);
+            return matchGroupInt >= getLower() && matchGroupInt <= getUpper();
```
Opeyem1a, line 55: `return getLower() <= matchGroupInt  && matchGroupInt <= getUpper();` looks a bit nicer imo

src/global/variables/clauses/IntegerLiteral.java
```diff
@@ -1,26 +1,81 @@
...
+        return lower;
     }
 
     public int getValue() {
         return value;
     }
 
+    @Override
+    public boolean validate(String matchGroupString) {
+        if (getLower() != getUpper()) {
```
Opeyem1a, line 53: Why check this condition? Let's just check if they are both defined

src/global/variables/clauses/IntegerLiteral.java
```diff
@@ -1,26 +1,81 @@
...
         this.value = value;
+        this.upper = value;
+        this.lower = value;
+
     }
 
     public IntegerLiteral(int value, String name) {
         super(name);
         this.value = value;
+        this.upper = value;
```
Opeyem1a, line 23: I don't think we should set `upper` and `lower` at all if they are not set by the user - this applies to the previous constructor as well

Opeyem1a (changes requested):

src/global/variables/clauses/IntegerLiteral.java
```diff
@@ -1,26 +1,108 @@
...
 
+import global.tools.InvalidClauseException;
+import global.utils.RegexUtil;
 import global.variables.Clause;
 
+import java.util.HashMap;
+import java.util.Map;
+
 public class IntegerLiteral extends Clause {
+    private Map<String, Boolean> MODES;
```
Opeyem1a, line 11: Woah this feels like a lot. I was imagining more like:
```java
class IntergerLiteral {
    private String mode;
    public static final RANGE = 'range';
    ...
}
```

then:
```java
this.modes = IntegerLiteral.RANGE;  // in a constructor
```

The map way feels like a lot for a value that can only be one of the modes at a time

src/global/variables/clauses/IntegerLiteral.java
```diff
@@ -1,26 +1,108 @@
...
+        return lower;
+    }
+
     public int getValue() {
         return value;
     }
 
+    @Override
+    public boolean validate(String matchGroupString) {
+        if (this.MODES.get("RANGE") != null && this.MODES.get("RANGE")) {
```
Opeyem1a, line 80: In the other implementation this could be ```if(this.mode == IntegerLiteral.RANGE)```

Opeyem1a (changes requested):

src/global/variables/clauses/IntegerLiteral.java
aadiraju: Yeah my bad I think I tried to implement an approach like TypeScript, but this makes sense.

Opeyem1a (approved):

PR #39, opened 2022-02-23
Methods modular programs hard q1

SamamthaH:


src/Methods/Modular_Programming/Hard/Q1/RightPrimeTest.java
```diff
@@ -0,0 +1,45 @@
...
+        assertTrue(RightPrime.isRightPrime(59393), "Your program does not handle large right truncatable primes");
+    }
+
+    int[] in = {3, 67, 345};
+    boolean [] boo = {true, false, false};
+    @RepeatedTest(3)
+    public void variousInputsTest(RepetitionInfo n){
+        boolean b = RightPrime.isRightPrime(in[n.getCurrentRepetition()-1]);
+        if(boo[n.getCurrentRepetition()-1])
+            assertTrue(b);
```
Hedgemon4, line 41: You should add an error message in the assertTrue and assertFalse statements here

Hedgemon4 (changes requested):

src/Methods/Modular_Programming/Hard/Q1/RightPrimeTest.java
SamamthaH: I have added some, I think they're good now but let me know

src/Methods/Modular_Programming/Hard/Q1/RightPrime.java
```diff
@@ -0,0 +1,42 @@
+package Methods.Modular_Programming.Hard.Q1;
+import java.util.Scanner;
+
+/*
+Write a program that will ask the user for a number and determine if it is a "right-truncatable prime",
+which is a special type of number where after removing each digit from the right, the resulting number
```
Hedgemon4, line 6: This question text was updated on the Junits examples document, but it does not look like the changes made it to the main document, and probably not onto the website.

src/Methods/Modular_Programming/Hard/Q1/RightPrimeTest.java
```diff
@@ -0,0 +1,45 @@
...
+import org.junit.jupiter.api.RepeatedTest;
+import org.junit.jupiter.api.RepetitionInfo;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class RightPrimeTest extends BaseTest{
+
+    public Clause[] testSentence() {
+        provideInput("5");
```
Hedgemon4, line 13: I would just use the example number here, but it is up to you

SanaBob, line 13: initialize input outside as int/String input = 5 and then put it into provide input.

src/Methods/Modular_Programming/Hard/Q1/RightPrimeTest.java
```diff
@@ -0,0 +1,45 @@
...
+    public void notPrimeAfterTruncationsTest(){
+        assertFalse(RightPrime.isRightPrime(1277), "Your program does not check the prime after multiple truncations");
+    }
+
+    @Test
+    public void largePrimeTest() {
+        assertTrue(RightPrime.isRightPrime(59393), "Your program does not handle large right truncatable primes");
+    }
+
+    int[] in = {3, 67, 345};
```
Opeyem1a, line 35: I feel like https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests could help here

Opeyem1a: Also let's name this variable better, `in` is a bit confusing

src/Methods/Modular_Programming/Hard/Q1/RightPrimeTest.java
```diff
@@ -0,0 +1,45 @@
...
+    boolean [] boo = {true, false, false};
```
Opeyem1a, line 36: `boo` is a tremendously cute name, but also a bit confusing

src/Methods/Modular_Programming/Hard/Q1/RightPrimeTest.java
```diff
@@ -0,0 +1,45 @@
...
+    @RepeatedTest(3)
+    public void variousInputsTest(RepetitionInfo n){
+        boolean b = RightPrime.isRightPrime(in[n.getCurrentRepetition()-1]);
+        if(boo[n.getCurrentRepetition()-1])
+            assertTrue(b, "Asserted false for a true random number");
```
Opeyem1a, line 41: This error message wouldn't be very helpful for a student. Perhaps something more like "Your method isRightPrime returned 'false' for an integer that is indeed a right-truncate-prime"

Opeyem1a (changes requested):

src/Methods/Modular_Programming/Hard/Q1/RightPrime.java
SamamthaH: I just double-checked the Junit examples doc and the text looks the same to me. What are you seeing that's different?

src/Methods/Modular_Programming/Hard/Q1/RightPrimeTest.java
SamamthaH: I try not to use example numbers in case of hard coding

SamamthaH: Why? I don't use it anywhere else I don't think.

Opeyem1a: I'd agree with Quan here, it's not used elsewhere but eventually you'll need to be providing an array of inputs instead of just one so the structure will look more like having a class variable defined at the top

SamamthaH: Sorry, can you reword that, I don't think I'm quite understanding what you mean. When you say I'll eventually need to provide an array of inputs, do you mean in future tests? or later in this test? It doesn't have the same format as my later inputs now that they're parameterized, and I don't see the point of adding unnecessary code

src/Methods/Modular_Programming/Hard/Q1/RightPrime.java
Opeyem1a: @Hedgemon4 Do you still think it's different?

src/Methods/Modular_Programming/Hard/Q1/RightPrimeTest.java
Opeyem1a: To reword, I think defining a `String input = "5";` outside of the testSentence() method is the move as it will be easier to switch to the `InputTest` framework Abhi and I make later

SamamthaH: okay, thanks for the clairifcation

src/Methods/Modular_Programming/Hard/Q1/MainTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class MainTest extends BaseTest{
+    String input = "5";
+
+    public Clause[] testSentence() {
+        provideInput(input);
```
Opeyem1a, line 14: Don't we need to put this inside `runMain()`, does this work? It's kinda cool if it does but for now still move it. Also let's move this test off the priority label since it needs input test

SamamthaH: nooooo it doesn't? I haven't run into any issues writing tests for it. Also what needs to get put into runMain()? The String input? or the provideInput? I didn't put the provideInput into runMain because then it would provide that input any other time I wanted to run the program, even when I didn't want it to provide that specific input.

src/Methods/Modular_Programming/Hard/Q1/RightPrime.java
Hedgemon4: Yeah, this is the same text in the examples document, but not the same as the gamification document

Hedgemon4: Main doucment says: 
[J] Write a program that will ask the user for a number and determine if it is a "right-truncatable prime", which is a special type of number where after removing each digit from the right, the resulting number after each digit removal is a prime number.
Eg: 2393 (prime)
239 ( Remove 3, still prime)
23 (remove 9, still prime)
2 (remove 3, still prime)
Make sure to use at least one java method in your solution.

Hedgemon4 (changes requested):
Gamification document text is still incorrect

src/Methods/Modular_Programming/Hard/Q1/RightPrime.java
SamamthaH: So you're saying I should update it in the gamification document? We've been making changes to the questions in the examples document to make them better questions for a while now, the current version I have is likely to be the correct one.

src/Methods/Modular_Programming/Hard/Q1/MainTest.java
SamamthaH: @Opeyem1a Why does this require inputTesting? I don't have any other plans to add more tests and all the tests function at the moment.

src/Methods/Modular_Programming/Hard/Q1/MainTest.java
```diff
@@ -0,0 +1,56 @@
...
+public class MainTest extends BaseTest{
+
+    public Clause[] testSentence() {
+        TestOption.isInputTest = true;
+        TestOption.defaultInput = "5";
+
+        return new Clause[]{
+                new StringLiteral("Enter a number"),
+                new NewLine(),
+                new StringLiteral("true"),
```
Hedgemon4, line 25: Why is this always true?

Hedgemon4 (changes requested):

src/Methods/Modular_Programming/Hard/Q1/MainTest.java
SamamthaH: Because this program uses a method to do the calculations, I only really run the clause for the default input, which should return true. So it doesn't need to be anything but true.

src/Methods/Modular_Programming/Hard/Q1/MainTest.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    static Stream<Arguments> inputProvider() {
+        return Stream.of(Arguments.of(3, true), Arguments.of(67, false), Arguments.of(345, false));
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    public void variousInputsTest(int inputs, boolean results){
+        boolean b = RightPrime.isRightPrime(inputs);
```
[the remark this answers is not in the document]
SamamthaH: That's the nice thing is that I can just test the method and only test the main program once for formatting. So StringLiteral doesn't need to change.

src/Methods/Modular_Programming/Hard/Q1/MainTest.java
```diff
@@ -0,0 +1,56 @@
+package Methods.Modular_Programming.Hard.Q1;
```
Opeyem1a, line 1: package names should all be lowercase

Opeyem1a (changes requested):

src/Methods/Modular_Programming/Hard/Q1/MainTest.java
Opeyem1a: Wait the package name is still incorrectly named? For clarity, I mean the packages, like the folders that this file is in

Opeyem1a (changes requested):

SamamthaH:
Sorry, the "renamed branch" thing was a mistype that I didn't realize I had done until it was already committed. I tried to rename the branch with the refactor but I didn't check if it took which is why it was still the same.

Opeyem1a (approved):
Looks good to me 👍🏾

aadiraju (approved):
Looks Good!

PR #40, opened 2022-02-23
Wrote CountAr Test and made it's package accordingly

SamamthaH:


src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/CountArTest.java
```diff
@@ -0,0 +1,73 @@
...
+                new NewLine(),
+                new IntegerLiteral(1),
+                new NewLine(),
+                new IntegerLiteral(0),
+                new NewLine(),
+                new IntegerLiteral(2),
+                new NewLine(),
+                new IntegerLiteral(0),
+                new NewLine(),
+                new IntegerLiteral(3),
```
SanaBob, line 33: try looping this as our previous one, might save time hardcoding this. Also the variables for IntegerLiteral should be defined outside of clause as int[] or multiple int and pasted into IntegerLiteral.

Hedgemon4: Yeah, I would just change this to a loop, but otherwise looks good

Hedgemon4 (changes requested):

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/CountArTest.java
```diff
@@ -0,0 +1,73 @@
...
+        int[] expected = new int[10];
+        int[] passed = new int[10];
+        for(int x = 0; x <10; x++){
+            int n = (int)(Math.random()*10);
+            passed[x] = n;
+            expected[n]++;
+        }
+        int[] actual = CountAr.countThisPlease(passed);
+
+        assertArrayEquals(expected, actual, "Expected output: " + Arrays.toString(expected) + " Received output: " + Arrays.toString(actual));
```
Opeyem1a, line 52: This error message would probably be confusing for a student to interpret what was actually wrong. Perhaps "Your method did not correctly count for the array [the array]".

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/CountArTest.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Test
+    public void methodOutOfBoundsException(){
+
+        try{
+            //noinspection ResultOfMethodCallIgnored
```
Opeyem1a, line 59: What does this mean?

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/CountArTest.java
```diff
@@ -0,0 +1,73 @@
...
+            CountAr.countThisPlease(new int[]{10, 10, 10, 9, 2,3,4,5,1,7, 5});
```
Opeyem1a, line 60: Run `ctrl` + `alt` + `l` on this file

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/CountArTest.java
```diff
@@ -0,0 +1,73 @@
...
+            fail("Returning array is larger than it should be");
```
Opeyem1a, line 61: Perhaps "Your method does not correctly handle the case where numbers larger than 9 are provided as input" or something - maybe run this by everyone to make sure it makes sense I'm just spitballing

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/CountArTest.java
```diff
@@ -0,0 +1,73 @@
...
+        }
+        catch (IndexOutOfBoundsException e){
+            assertTrue(true);
```
Opeyem1a, line 64: You can just do nothing here, we already know we're gucci if it reaches this line so no need for an assert

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/CountArTest.java
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+    @Test
+    public void methodPassingAnArrayLargerThan10(){
+        int[] expected = {1, 2, 3, 0, 1, 1, 1, 1, 0, 1};
+        int[] actual = CountAr.countThisPlease(new int[]{2, 2, 4, 2, 1, 9, 5, 1, 6, 0, 7});
+        assertArrayEquals(expected, actual, "Your program did not count all the passed numbers");
```
Opeyem1a, line 71: "...did not correctly count all numbers given in the input"

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/CountArTest.java
```diff
@@ -0,0 +1,73 @@
...
+            fail("Returning array is larger than it should be");
+        }
+        catch (IndexOutOfBoundsException e){
+            assertTrue(true);
+        }
+    }
+    @Test
+    public void methodPassingAnArrayLargerThan10(){
+        int[] expected = {1, 2, 3, 0, 1, 1, 1, 1, 0, 1};
+        int[] actual = CountAr.countThisPlease(new int[]{2, 2, 4, 2, 1, 9, 5, 1, 6, 0, 7});
```
Opeyem1a, line 70: You could probably randomly generate just like before and calculate what expected should be instead of hardcoding this case

Opeyem1a (changes requested):

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/CountArTest.java
SamamthaH: does a test auto pass if it gets to the end without an assert?

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/CountArTest.java
SamamthaH: It's suppression for the ResultOfMethodCallIgnored warning, if you want I can get rid of it. I don't think it likes that I don't store the returned array.

Opeyem1a: I think it's just warning you that the returned value of the method isn't stored, you should be able to just save it to a variable to avoid the warning, or just ignore the warning altogether

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/CountArTest.java
Opeyem1a: The concern is just that having this be hard-coded makes cheating around it easier. Designing robust tests should make cheating near-impossible. The added time is maybe seconds, so it's not too concerning, but if that's a worry we could also hardcode more cases into this test - a test checking for how the method runs with arrays larger than 10 should probably check bigger arrays in general and not just an array of length 11

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/CountAr.java
```diff
@@ -0,0 +1,28 @@
...
+
+/*
+Write a program that defines (in the main method) an array of positive numbers in the range of 0 and 9 with the following values :
+[ 1, 3, 4, 5, 4, 9, 4, 9, 9, 7, 7, 3 ].
+This program will also have a method called countThisPlease() that counts the frequency of each number in that array.
+With the input array you've defined, the method would count that there are no 0's, one 1's, no 2's, two 3's, three 4's, etc.,
+and store those counts into a new array [ 0, 1, 0, 2, 3, 1, 0, 2, 0, 3] for each possible value between 0 to 9.
+That new array of counts will be returned as the method's output, which the main method should print out,
+with one element on each line.
+ */
```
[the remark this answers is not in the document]
SamamthaH: @Opeyem1a I'm pretty sure the description on the JUnit examples document is more accurate, should I update the gamification document accordingly?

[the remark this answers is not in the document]
aadiraju: Yes please do update the gamification document with this updated example

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/MainTest.java
```diff
@@ -0,0 +1,65 @@
...
+    public void methodPassingAnArrayLargerThan10() {
+        int[] expected = new int[10];
+        int[] passed = new int[30];
+        for (int x = 0; x < passed.length; x++) {
+            int n = (int) (Math.random() * 10);
+            passed[x] = n;
+            expected[n]++;
+        }
+        int[] actual = CountAr.countThisPlease(passed);
+        assertArrayEquals(expected, actual, "Your program did not correctly count all the numbers given in the input");
```
aadiraju, line 63: Maybe make this a tad more specific like `Your program should only count arrays that have exactly 10 values` then you can even check for arrays that have less than 10 elements.

SamamthaH: I think there's been a slight misunderstanding, it's not supposed to fail if it counts more than 10 values. The program should be able to take an array of any length as long as the elements fall in the range of 0-9. I can write another test that checks under 10 elements doesn't crash.

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/MainTest.java
```diff
@@ -0,0 +1,65 @@
...
+        int[] expected = new int[10];
+        int[] passed = new int[10];
+        for (int x = 0; x < 10; x++) {
+            int n = (int) (Math.random() * 10);
+            passed[x] = n;
+            expected[n]++;
+        }
+        int[] actual = CountAr.countThisPlease(passed);
+
+        assertArrayEquals(expected, actual, "Your method did not correctly count for the array: " + Arrays.toString(passed));
```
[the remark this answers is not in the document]
SamamthaH: I mean, I'm pretty sure I did that. That's what the Arrays.toString(passed) is

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/MainTest.java
aadiraju: Oh my bad then. Try to check that the array is definitely 10 elements or over then, and fail otherwise. You might have to update the question to ensure that this requirement is communicated as well

SamamthaH: What exactly is the requirement? Are you thinking of adding a condition to what the array input length must be?

aadiraju: Sorry about the misunderstanding. Upon looking at the question again I think your tests are fine. But I think you need to add the question type to the code? I'm not sure if this is a parsons or a Java

SamamthaH: I thought it was assumed if there wasn't a parsons label that it was java

aadiraju: Yep you're right. I wrote this before the meeting so I was unsure.

aadiraju (approved):
It looks good to me, the question is worded a little weirder so that might need to be changed. What do you think @Opeyem1a ?

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/CountAr.java
```diff
@@ -0,0 +1,28 @@
+package Arrays.Programs_involving_Data_Sequences.Hard.Q1;
```
Opeyem1a, line 1: package names should be lowercase

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/MainTest.java
```diff
@@ -0,0 +1,65 @@
...
+import global.BaseTest;
+import global.variables.*;
+import org.junit.jupiter.api.RepeatedTest;
+import org.junit.jupiter.api.Test;
+
+import java.util.Arrays;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class MainTest extends BaseTest {
```
Opeyem1a, line 12: Is this not a parsons question?

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/MainTest.java
```diff
@@ -0,0 +1,65 @@
...
+    int[] inputs = {0, 1, 0, 2, 3, 1, 0, 2, 0};
```
Opeyem1a, line 13: is this variable named accurately? I don;t understand why this is an input? Is this supposed to be the correct counts?

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/MainTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+    public Clause[] testSentence() {
+        Clause[] output = new Clause[19];
+        for (int x = 0; x < 17; x += 2) {
```
Opeyem1a, line 17: A while loop would make this easier to read

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/MainTest.java
```diff
@@ -0,0 +1,65 @@
...
+            output[x] = new IntegerLiteral(inputs[x / 2]);
+            output[x + 1] = new NewLine();
+        }
+        output[18] = new IntegerLiteral(3);
```
Opeyem1a, line 21: Does `output` mean the test sentence? These variable names make it tough to interpret what's supposed to be happening here

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/MainTest.java
```diff
@@ -0,0 +1,65 @@
...
+    @RepeatedTest(50)
+    public void methodTestRandomNums() {
+        int[] expected = new int[10];
+        int[] passed = new int[10];
+        for (int x = 0; x < 10; x++) {
+            int n = (int) (Math.random() * 10);
+            passed[x] = n;
+            expected[n]++;
+        }
+        int[] actual = CountAr.countThisPlease(passed);
```
Opeyem1a, line 38: You won't have access to this method when you put this in the website

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/MainTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+        assertArrayEquals(expected, actual, "Your method did not correctly count for the array: " + Arrays.toString(passed));
+    }
+
+    @Test
+    public void methodOutOfBoundsException() {
```
Opeyem1a, line 44: Not sure what this is testing

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/MainTest.java
```diff
@@ -0,0 +1,65 @@
...
+        output[18] = new IntegerLiteral(3);
+        return output;
+    }
+
+    public void runMain() {
+        CountAr.main(new String[]{});
+    }
+
+    @RepeatedTest(50)
+    public void methodTestRandomNums() {
```
Opeyem1a, line 30: After looking at this file I think the better move is to use inputProvider and hardcode a series of input arrays and the correct output counts array for them

Opeyem1a (changes requested):

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/MainTest.java
Opeyem1a: We need to add support for checking method headers to ensure patterns like this have the correct argument and return type to be used in this way

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/MainTest.java
SamamthaH: nope

src/Arrays/Programs_involving_Data_Sequences/Hard/Q1/MainTest.java
SamamthaH: The question asks for the user to count an array with number ranges from 0-9, this is supposed to test that it's only in that range and not outside it, but if you think that's unnecessary I'll get rid of it.

Hedgemon4:
This question has been redone in https://github.com/canvas-gamification/canvas-gamification-junit-tests/pull/263

PR #41, opened 2022-02-24
pre_defined_classes-mathematical_programs-hard-q1

Hedgemon4:


SamamthaH (changes requested):
I think this is labeled wrong, I'm pretty sure this is pre-defined classes mathematical programs hard q1
Also not entirely sure why ProphetRobot is being included in this

Hedgemon4:
> I think this is labeled wrong, I'm pretty sure this is pre-defined classes mathematical programs hard q1 Also not entirely sure why ProphetRobot is being included in this

It is included because I was playing around with something, and testing some values with it. It should not actually modify anything.

Hedgemon4:
> I think this is labeled wrong, I'm pretty sure this is pre-defined classes mathematical programs hard q1 Also not entirely sure why ProphetRobot is being included in this

This is also the same name which is on the Junit example document, so if it it wrong there then it is mislabeled, I will double-check on the excel sheet

SamamthaH (approved):

src/pre_defined_classes/mathematical_program/hard/q1/JavademicTest.java
```diff
@@ -0,0 +1,23 @@
...
+import pre_defined_classes.mathematical_program.hard.q1.Javademic;
+
+public class JavademicTest extends BaseTest {
+
+    public Clause[] testSentence() {
+        return new Clause[]{
+                new StringLiteral("Num cases of infections in "),
+                new IntegerLiteral(30),
+                new StringLiteral(" days is "),
+                new StringLiteral("13099.97821824974"),
```
Opeyem1a, line 16: I think this should also use DoubleLiteral so we can validate the math, but it's fine for now - make an issue so we can revisit this later

Opeyem1a (approved):

src/pre_defined_classes/mathematical_program/hard/q1/JavademicTest.java
Hedgemon4: Yeah, it definitely should, but I was just working with what we have for now.

src/pre_defined_classes/mathematical_program/hard/q1/JavademicTest.java
```diff
@@ -0,0 +1,34 @@
...
+    final double xtExpected = 13099.97821824974;
+
+    public Clause[] testSentence() {
+        return new Clause[]{
+                new StringLiteral("Num cases of infections in "),
+                new IntegerLiteral(30),
+                new StringLiteral(" days is "),
+                //These are the possible range of values if they select on of the parsons distractors, and the proper
+                //value is tested for below
+                new DoubleLiteral(44, 1.23453340530816128E17, "xt"),
```
Opeyem1a, line 21: Use `Double.MAX_VALUE` for upper boundaries like this

Opeyem1a (changes requested):

src/pre_defined_classes/mathematical_program/hard/q1/JavademicTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import global.BaseTest;
+import global.variables.Clause;
+import global.variables.clauses.DoubleLiteral;
+import global.variables.clauses.IntegerLiteral;
+import global.variables.clauses.StringLiteral;
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class JavademicTest extends BaseTest {
```
Opeyem1a, line 11: Rename to `MainTest` and comment if Parson's

Opeyem1a (changes requested):

SanaBob (dismissed):

SamamthaH (approved):

aadiraju (dismissed):
I think this looks fine and can be merged. I will let @Opeyem1a take a look

src/pre_defined_classes/mathematical_program/hard/q1/MainTest.java
```diff
@@ -0,0 +1,33 @@
...
+import global.BaseTest;
+import global.variables.Clause;
+import global.variables.clauses.DoubleLiteral;
+import global.variables.clauses.IntegerLiteral;
+import global.variables.clauses.StringLiteral;
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MainTest extends BaseTest {
+    // Parsons (with distractors)
```
Opeyem1a, line 12: is this the way we decided on for these? I feel like I see it differently, but I'll leave the final call on that up to @aadiraju

Opeyem1a (changes requested):

src/pre_defined_classes/mathematical_program/hard/q1/MainTest.java
aadiraju: I think we decided not to have brackets so this would be `// Parsons with Distractors` instead.

src/pre_defined_classes/mathematical_program/hard/q1/MainTest.java
```diff
@@ -0,0 +1,35 @@
...
+import global.variables.clauses.IntegerLiteral;
+import global.variables.clauses.StringLiteral;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MainTest extends BaseTest {
+    // Parsons with distractors
+
+    final double xtExpected = 13100.00;
```
SamamthaH, line 15: Why have this as a variable when it's just used to check the math?

Hedgemon4: I don't really think this is a problem

Opeyem1a (approved):

PR #45, opened 2022-02-24
pre defined classes mathematical programs hard Q2

SamamthaH:
Wrote clause for Q2, examined all distractors and the one that is not a compiler error is blatantly false so no other tests are needed.


src/pre_defined_classes/mathematical_program/hard/Q2/Javademic.java
```diff
@@ -0,0 +1,20 @@
+package pre_defined_classes.mathematical_program.hard.Q2;
```
Hedgemon4, line 1: package q2 should be lower case

src/pre_defined_classes/mathematical_program/hard/q2/JavademicTest.java
```diff
@@ -0,0 +1,20 @@
+package pre_defined_classes.mathematical_program.hard.q2;
+
+import global.BaseTest;
+import global.variables.*;
+
+public class JavademicTest extends BaseTest {
```
Opeyem1a, line 6: Rename to `MainTest` and comment if Parson's

src/pre_defined_classes/mathematical_program/hard/q2/JavademicTest.java
```diff
@@ -0,0 +1,20 @@
...
+    public Clause[] testSentence() {
+        return new Clause[]{
+                new StringLiteral("It will take "),
+                new IntegerLiteral(40),
+                new StringLiteral(" days to infect "),
+                new IntegerLiteral(1000000),
```
Opeyem1a, line 12: `DoubleLiteral`

Opeyem1a (changes requested):

src/pre_defined_classes/mathematical_program/hard/q2/JavademicTest.java
SamamthaH: Why double literal?

Hedgemon4: I think this is because some of the distractors return doubles

src/pre_defined_classes/mathematical_program/hard/q2/MainTest.java
```diff
@@ -0,0 +1,21 @@
+package pre_defined_classes.mathematical_program.hard.q2;
+
+import global.BaseTest;
+import global.variables.*;
+
+public class MainTest extends BaseTest {
+    //parsons
+    public Clause[] testSentence() {
```
Hedgemon4, line 8: You might want to make this less hardcoded, just catch the values, and then check the math to make sure they return the correct value type and give better feedback, because right now if they use a distractor, they will just be told the code does not match the proper syntax/structure.

Hedgemon4 (changes requested):

src/pre_defined_classes/mathematical_program/hard/q2/JavademicTest.java
SamamthaH: The only distractor that would change the output from an integer is double t = Math.exp( xt / x0 ) / Math.exp( 1 + r ); which actually causes the program to print "Infinity" so, is the recommendation that I change it to String instead?

src/pre_defined_classes/mathematical_program/hard/q2/MainTest.java
SamamthaH: Distractors:
 int xt = 1,000,000, x0 = 2; breaks
double t = Math.exp( xt / x0 ) / Math.exp( 1 + r ); Infinity
int t = Math.exp( xt / x0 ) / Math.exp( 1 + r ); breaks
int t = ( int )Math.exp( xt / x0 ) / Math.exp( 1 + r ); breaks
int t = ( double )( Math.exp( xt / x0 ) / Math.exp( 1 + r )); breaks
int t = Math.round( numerator / denominator ); breaks
double numerator = log( xt / x0 ); breaks
double denominator = log( 1 + r ); breaks

All but one of the distractors just breaks the program, there is no way for the math to be wrong except the infinity value, which still doesn't require a math test so I really don't see why I should make the math test

Hedgemon4: I don't disagree with that point here, I just think that it is different then we have decided to write the tests, and that it is better to stick to the same sort of pattern

Hedgemon4 (commented):
See my comment

Hedgemon4 (dismissed):
I just had one comment, but otherwise good job

src/pre_defined_classes/mathematical_program/hard/q2/MainTest.java
```diff
@@ -0,0 +1,29 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MainTest extends BaseTest {
+    // Parsons with distractors
+    public Clause[] testSentence() {
+        return new Clause[]{
+                new StringLiteral("It will take "),
+                new IntegerLiteral(40, "days"),
```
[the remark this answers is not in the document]
SamamthaH: It won't let me not input a value, but I still have the math test to give clearer feedback.

src/pre_defined_classes/mathematical_program/hard/q2/Javademic.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+public class Javademic {
+    public static void main(String[] args) {
+        double r = 0.39;
+        int xt = 1000000;
+        int x0 = 2;
+        double numerator = Math.log( xt / x0 );
+        double denominator = Math.log( 1 + r );
+        int t = ( int )Math.round( numerator / denominator );
```
Hedgemon4, line 27: We should add something to tell them to round this, because I think there might be a way to solve the question without the rounding, so if we want to do it this way we should say that

Hedgemon4 (changes requested):

src/pre_defined_classes/mathematical_program/hard/q2/Javademic.java
SamamthaH: do you mean a way with the distractors? Or just a way in general if they were able to code it?

src/pre_defined_classes/mathematical_program/hard/q2/MainTest.java
```diff
@@ -0,0 +1,32 @@
...
+        };
+    }
+
+    public void runMain() {
+        Javademic.main(new String[0]);
+    }
+
+    @Test
+    public void mathTest(){
+        assertEquals(getItemByName("days"), "40", "Your calculation for the number of days it takes to infect 1000000 is incorrect");
```
Opeyem1a, line 29: Just add periods `.` to the ends of these error messages and we're good to go!

Opeyem1a (changes requested):

Opeyem1a (approved):

PR #49, opened 2022-02-25
pre defined classes-mathematical program-medium-q3

Hedgemon4:
Need DoubleLiteral to finish


src/pre_defined_classes/mathematical_program/medium/q3/TrigIdentityTest.java
```diff
@@ -0,0 +1,40 @@
...
+                new StringLiteral("Their sum is\\: "),
+                new DoubleLiteral(0, 2.0, "sum")
+        };
+    }
+
+    public void runMain() {
+        TrigIdentity.main(new String[0]);
+    }
+
+    @Test
```
SamamthaH, line 34: Based on the fact that there are no distractors and that the way the program is set up, it won't compile if the math is out of order, they can't really get the math wrong. It might be more useful to say something like "Your first output value should be sin^2"

Hedgemon4: I don't know, Opey said that we should still check the math

Hedgemon4: But I could add a separate test, or we would provide the format output

Hedgemon4: And it actually won't pass the clause test if it is in the wrong order, so we just need to provide the output format

SamamthaH: I figured it might give them a better idea of what order to put them in, but it might not be necessary.

src/pre_defined_classes/mathematical_program/medium/q3/TrigIdentityTest.java
```diff
@@ -0,0 +1,40 @@
...
+    public void testDoubleValues(){
+        assertEquals(sin2 + "", getItemByName("sin2"), "Your value for sin^2 is incorrect");
```
Opeyem1a, line 36: All 3 of these `assertEquals` should have `delta` set so the value doesn't need to be exact exact (because of the way floating-point accuracy works)

src/pre_defined_classes/mathematical_program/medium/q3/TrigIdentityTest.java
```diff
@@ -0,0 +1,40 @@
...
+import global.variables.Clause;
+import global.variables.clauses.DoubleLiteral;
+import global.variables.clauses.NewLine;
+import global.variables.clauses.StringLiteral;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class TrigIdentityTest extends BaseTest {
+    final double cos2 = 0.2500000000000001;
```
Opeyem1a, line 13: Once you set deltas on your assertions later, you could probably have this just be `0.25` and `0.75` respectively

Opeyem1a (changes requested):

src/pre_defined_classes/mathematical_program/medium/q3/MainTest.java
```diff
@@ -0,0 +1,42 @@
...
+import global.variables.clauses.StringLiteral;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MainTest extends BaseTest {
+    // Parsons
+    final double cos2 = 0.25;
+    final double sin2 = 0.75;
+    final double sum = 1.0;
```
SamamthaH, line 16: Again, why put numbers you only use in the math test up here? Other than that it's good.

src/pre_defined_classes/mathematical_program/medium/q3/MainTest.java
```diff
@@ -0,0 +1,39 @@
...
+public class MainTest extends BaseTest {
+    // Parsons
+
+    public Clause[] testSentence() {
+        return new Clause[]{
+                new StringLiteral("The value of sin\\^2 is: "),
+                new DoubleLiteral("sin2"),
+                new NewLine(),
+                new StringLiteral("The value of cos\\^2 is: "),
+                new DoubleLiteral( "cos2"),
```
Opeyem1a, line 21: Format this file

Opeyem1a (changes requested):

Opeyem1a (approved):

PR #53, opened 2022-02-25
loops_simple_programs_with_repitition_easy_q6

SanaBob:
loops_simple_programs_with_repitition_easy_q6

### DIFF
--- a/src/loops/simple_programs_with_repitition/easy/CountItUp.java
+++ b/src/loops/simple_programs_with_repitition/easy/CountItUp.java
@@ -1,0 +1,25 @@
+package temp;
+
+/*
+Write a program that will ask the user for a sentence and then a character and 
+prints out the number of times the entered character occurs in the user's sentence.
+*/
+
+import java.util.Scanner;
+
+public class CountItUp {
+    public static void main(String[] args) {
+        Scanner input = new Scanner(System.in);
+        System.out.println("Enter a sentence: ");
+        String sent = input.nextLine();
+        sent.toLowerCase();
+        System.out.println("Enter the character to count: ");
+        char c = (char) (input.next().charAt(0));
+        int count = 0;
+        for (int i = 0; i < sent.length(); i++)
+            if (sent.charAt(i) == c)
+                count++;
+        System.out.println("The number of times " + c + " occurs in the sentence is "
+        + count + " times.");
+    }
+}
--- a/src/loops/simple_programs_with_repitition/easy/CountItUp.java
+++ b/src/loops/simple_programs_with_repitition/easy/CountItUp.java
@@ -1,4 +1,4 @@
-package temp;
+package loops.simple_programs_with_repitition.easy;
 
 /*
 Write a program that will ask the user for a sentence and then a character and 
--- a/src/loops/simple_programs_with_repitition/easy/q6/CountItUp.java
+++ b/src/loops/simple_programs_with_repitition/easy/q6/CountItUp.java
@@ -1,4 +1,4 @@
-package loops.simple_programs_with_repitition.easy;
+package loops.simple_programs_with_repitition.easy.q6;
 
 /*
 Write a program that will ask the user for a sentence and then a character and 
--- a/src/loops/simple_programs_with_repitition/easy/q6/CountItUp.java
+++ b/src/loops/simple_programs_with_repitition/easy/q6/CountItUp.java
@@ -1,8 +1,9 @@
 package loops.simple_programs_with_repitition.easy.q6;
 
 /*
-Write a program that will ask the user for a sentence and then a character and 
-prints out the number of times the entered character occurs in the user's sentence.
+Write a program that will ask the user for a sentence and then a character 
+and prints out the number of times the entered character occurs in the user's sentence. 
+Assume the input is a sentence and a character in lowercase.
 */
 
 import java.util.Scanner;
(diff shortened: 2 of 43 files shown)


src/loops/simple_programs_with_repitition/easy/CountItUp.java
```diff
@@ -0,0 +1,25 @@
+package loops.simple_programs_with_repitition.easy;
+
+/*
+Write a program that will ask the user for a sentence and then a character and
```
Hedgemon4, line 4: There are two problems. The first is that it is not specified in the question if we should ignore case or not, and the document code does, but yours does not. This should be corrected to avoid confusion. Also, again there is no sample output, so they can't actually match what your clause statement is testing for.

Hedgemon4 (changes requested):

src/loops/simple_programs_with_repitition/easy/CountItUp.java
Hedgemon4: I also just looked and saw that these are Parson questions, so technically they are not writing the code themselves, but I still feel a sample output would be useful and is something we should talk about in a meeting.

Hedgemon4: I think we should update this question to specify if we want to ignore Case or not, because the solution code does, but the code you wrote in your test does not.

src/loops/simple_programs_with_repitition/easy/q6/CountItUpTest.java
```diff
@@ -0,0 +1,38 @@
...
+import global.BaseTest;
+
+public class CountItUpTest extends BaseTest {
+    
+    String input1 = "This is just for testing. Do not copy!";
+    String input2 = "s";
+
+    public Clause[] testSentence() {
+
+        Clause[] output = new Clause[5];
```
Opeyem1a, line 14: Return the clause array the normal way

Opeyem1a (changes requested):

src/loops/simple_programs_with_repitition/easy/q6/CountItUpTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+import global.variables.*;
+import global.variables.clauses.*;
+import global.BaseTest;
+
+public class CountItUpTest extends BaseTest {
+    
+    String input1 = "This is just for testing. Do not copy!";
+    String input2 = "s";
+
```
SamamthaH, line 11: I would pick a different letter than s, or put a capital s in the test string. As it is, if they misplace the .toLowerCase() it would still pass the tests.

SanaBob: this should be resolved when inputTest is implemented

src/loops/simple_programs_with_repitition/easy/q6/MainTest.java
```diff
@@ -0,0 +1,60 @@
...
+            new IntegerLiteral("numberOutput"),
+            new StringLiteral(" times.")
+        };
+    }
+
+    public void runMain() {
+        CountItUp.main(new String[0]);
+    }
+
+    public static int countingLetter(String message, char c) {
```
Opeyem1a, line 39: This helper method also isn't needed, just hardcode the output

src/loops/simple_programs_with_repitition/easy/q6/MainTest.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void testWithInput(String input1, String input2) throws InvalidClauseException{
+        char newInput2 = input2.charAt(0);
+        runWithInput(input1 + System.lineSeparator() + input2, new Clause[]{
+            new StringLiteral(input2)
+        });
+        assertEquals(Integer.parseInt(getItemByName("numberOutput")), countingLetter(input1, newInput2), "Calculated value of output is incorrect");
```
Opeyem1a, line 58: This error message needs improvement

Opeyem1a (changes requested):

src/loops/simple_programs_with_repitition/easy/q6/CountItUp.java
```diff
@@ -0,0 +1,25 @@
+package loops.simple_programs_with_repitition.easy.q6;
+
+/*
+Write a program that will ask the user for a sentence and then a character and
```
Opeyem1a, line 4: This question needs to explain whether upper and lower case of the same character should count as the same - I added the bad question label for you to bring this up in question report later

Opeyem1a (changes requested):

src/loops/simple_programs_with_repitition/easy/q6/MainTest.java
```diff
@@ -0,0 +1,51 @@
...
+        CountItUp.main(new String[0]);
+    }
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of("This is just for testing. Do not copy!", "s", 4), Arguments.of("This is for second testing. Do not copy!", "o", 5), Arguments.of("NNNNNNNN", "n", 0), Arguments.of(" ", "z", 0));
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void testWithInput(String input1, String input2, int output) throws InvalidClauseException{
```
Opeyem1a, line 45: Variables could be named better but it's alright

Opeyem1a (commented):
The code is okay, but we need to confirm if the bad question is resolved both here and in the website

Opeyem1a (approved):

PR #54, opened 2022-02-25
loops_simple_programs_with_repitition_easy_q7

SanaBob:
loops_simple_programs_with_repitition_easy_q7


Hedgemon4 (approved):
Looks good

src/loops/simple_programs_with_repitition/easy/q7/AlTeRnAtInGcAsETest.java
```diff
@@ -0,0 +1,41 @@
...
+        return output;
+
+    }
+
+    public void runMain() {
+        provideInput(input);
+        AlTeRnAtInGcAsE.main(new String[0]);
+    }
+
+    public static String alternate(String sentence) {
```
SamamthaH, line 28: Since you are providing the input, can't you alternate the case on your own and put it in the output without needing this method?

src/loops/simple_programs_with_repitition/easy/q7/AlTeRnAtInGcAsETest.java
```diff
@@ -0,0 +1,41 @@
...
+import global.variables.clauses.*;
+import global.BaseTest;
+
+public class AlTeRnAtInGcAsETest extends BaseTest {
+
+    String input = "This is a sentence. This is another sentence.";
+
+    public Clause[] testSentence() {
+
+        Clause[] output = new Clause[3];
```
Opeyem1a, line 13: Return the clause array the normal way

Opeyem1a (changes requested):

Hedgemon4 (changes requested):
This should be merging into Beta

src/loops/simple_programs_with_repitition/easy/q7/MainTest.java
```diff
@@ -0,0 +1,60 @@
...
+        int i = 0;
+        while (i < sentence.length()) {
+            if (i % 2 == 0)
+                result += Character.toUpperCase(sentence.charAt(i));
+            else
+                result += Character.toLowerCase(sentence.charAt(i));
+            i++;
+        }
+        return result;
+    }
```
SamamthaH, line 45: Since you know what values you are inputting, I don't think you need this method, you can just figure out what the correct output should be and put it as an argument for the parameterized test to use in the assertEquals

src/loops/simple_programs_with_repitition/easy/q7/MainTest.java
```diff
@@ -0,0 +1,60 @@
...
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void testWithInput(String input) throws InvalidClauseException{
+        runWithInput(input, new Clause[] {
+            new StringLiteral(alternate(input), "stringOutput")
+        });
+        assertEquals(getItemByName("stringOutput"), alternate(input), "Your output string output is not correct");
+    }
+
+}
```
Opeyem1a, line 60: EOF newline

src/loops/simple_programs_with_repitition/easy/q7/MainTest.java
```diff
@@ -0,0 +1,60 @@
...
+                result += Character.toUpperCase(sentence.charAt(i));
+            else
+                result += Character.toLowerCase(sentence.charAt(i));
+            i++;
+        }
+        return result;
+    }
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of("This is just for testing. Do not copy!"), Arguments.of("This is second test"), Arguments.of(" ")); // Doesn't work with empty string
```
Opeyem1a, line 48: If it doesn't work with empty string, either it's not needed in tests or we need to change the question to explain this - I've marked this as bad question as well due to this

src/loops/simple_programs_with_repitition/easy/q7/MainTest.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void testWithInput(String input) throws InvalidClauseException{
+        runWithInput(input, new Clause[] {
+            new StringLiteral(alternate(input), "stringOutput")
+        });
+        assertEquals(getItemByName("stringOutput"), alternate(input), "Your output string output is not correct");
```
Opeyem1a, line 57: Error message needs heavy work

Opeyem1a (changes requested):

src/loops/simple_programs_with_repitition/easy/q7/MainTest.java
```diff
@@ -0,0 +1,60 @@
...
+                result += Character.toUpperCase(sentence.charAt(i));
+            else
+                result += Character.toLowerCase(sentence.charAt(i));
+            i++;
+        }
+        return result;
+    }
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of("This is just for testing. Do not copy!", alternate("This is just for testing. Do not copy!")), Arguments.of("This is second test", "ThIs iS SeCoNd tEsT"), Arguments.of("THIS IS ANOTHER HARDCODED TEST", "ThIs iS AnOtHeR HaRdCoDeD TeSt"), Arguments.of(" ", " ")); // Doesn't work with empty string
```
Opeyem1a, line 48: Wait not like this, either hardcode the string as the second parameter of the `.of()` as a string (i.e. without the helper method) or use the helper method only in the `testWithInput()`, doing both is redundant since you're injecting a value using the helper method and then checking if it equals another value computed with the helper method. The problem being that if the helper method is ever wrong, the test will still pass since it only depends on itself

Opeyem1a (changes requested):

src/loops/simple_programs_with_repitition/easy/q7/MainTest.java
```diff
@@ -0,0 +1,45 @@
...
+                new PlaceHolder()
+        };
+    }
+
+    public void runMain() {
+        AlTeRnAtInGcAsE.main(new String[0]);
+    }
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of("This is first test!", "ThIs iS FiRsT TeSt!"), Arguments.of("This is second test", "ThIs iS SeCoNd tEsT"), Arguments.of("THIS IS ANOTHER HARDCODED TEST", "ThIs iS AnOtHeR HaRdCoDeD TeSt"), Arguments.of(" ", " ")); // Doesn't work with empty string
```
Opeyem1a, line 35: Does it not work with empty string? It seems like it works, wouldn't it just return empty string too?

Opeyem1a: Otherwise this is good though

Opeyem1a (changes requested):

Opeyem1a (approved):

PR #55, opened 2022-02-25
loops_simple_programs_with_repitition_easy_q8

SanaBob:
loops_simple_programs_with_repitition_easy_q8


Hedgemon4:
You could also add the test case provided, but that is more up to you. Otherwise, this looks good.

Hedgemon4 (dismissed):

src/loops/simple_programs_with_repitition/easy/q8/CaesarSikeTest.java
```diff
@@ -0,0 +1,40 @@
...
+        return output;
+        
+    }
+
+    public void runMain() {
+        provideInput(input1 + System.lineSeparator() + input2);
+        CaesarSike.main(new String[0]);
+    }
+
+    public static String encrypt(String sentence, int offset) {
```
SamamthaH, line 31: Same as previous comments. I believe the method is unnecessary since you dictate the input.

src/loops/simple_programs_with_repitition/easy/q8/CaesarSikeTest.java
```diff
@@ -0,0 +1,40 @@
...
+import global.BaseTest;
+
+public class CaesarSikeTest extends BaseTest {
+    
+    String input1 = "This is a sentence. This is another sentence.";
+    int input2 = 2; 
+
+    public Clause[] testSentence() {
+
+        Clause[] output = new Clause[5];
```
Opeyem1a, line 14: return this the normal way

src/loops/simple_programs_with_repitition/easy/q8/CaesarSikeTest.java
Opeyem1a: This is a bit more nuanced since it's easy to make mistakes doing something that finicky. A util method is alright here, but we should have some manual tests too where we explicitly define the input and encrypted version

Opeyem1a (changes requested):

src/loops/simple_programs_with_repitition/easy/q8/MainTest.java
```diff
@@ -0,0 +1,58 @@
...
+        int i = 0;
+        while (i < sentence.length()) {
+            result += (char)(sentence.charAt(i) + offset);
+            i++;
+        }
+        return result;
+    }
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of("This is just for testing. Do not copy!", 1), Arguments.of("This is second test", 100), Arguments.of(" ", 0)); // Doesn't work with empty string
```
Opeyem1a, line 48: What about negative offsets? Also the same comment about empty strings not working - that is an important thing to raise when you see it as either the question or the test needs to change.

I'm adding the bad question tag for this reason

src/loops/simple_programs_with_repitition/easy/q8/MainTest.java
```diff
@@ -0,0 +1,58 @@
...
+            new PlaceHolder()
+        };
+        
+    }
+
+    public void runMain() {
+        CaesarSike.main(new String[0]);
+    }
+
+    public static String encrypt(String sentence, int offset) {
```
Opeyem1a, line 37: The use of a helper, in this case, is a bit more justifiable, so it's okay here

Opeyem1a (changes requested):

src/loops/simple_programs_with_repitition/easy/q8/MainTest.java
```diff
@@ -0,0 +1,58 @@
...
+        int i = 0;
+        while (i < sentence.length()) {
+            result += (char)(sentence.charAt(i) + offset);
+            i++;
+        }
+        return result;
+    }
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of("This is just for testing. Do not copy!", 1, encrypt("This is just for testing. Do not copy!", 1)), Arguments.of("This is second test", 100, encrypt("This is second test", 100)), Arguments.of("hard", 1, "ibse"), Arguments.of(" ", 1, "!")); // Doesn't work with empty string
```
Opeyem1a, line 48: Purely for readability, maybe put each `Arguments.of()` on a new line

Opeyem1a (changes requested):

SanaBob:
I'm doing a mix of helper and hardcoded here, is it okay? or change it to hardcoded only? (would be hard to create them though)

src/loops/simple_programs_with_repitition/easy/q8/MainTest.java
```diff
@@ -0,0 +1,63 @@
...
+        return result;
+    }
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(
+            Arguments.of("This is just for testing. Do not copy!", 1, encrypt("This is just for testing. Do not copy!", 1)),
+            Arguments.of("This is second test", 100, encrypt("This is second test", 100)),
+            Arguments.of("hard", 1, "ibse"),
+            Arguments.of(" ", 1, "!")
+        ); // Doesn't work with empty string
```
Opeyem1a, line 53: Wait wouldn't empty string just return empty string regardless of offset? If it doesn't why not

Opeyem1a (changes requested):

src/loops/simple_programs_with_repitition/easy/q8/MainTest.java
SanaBob: space is also considered a character in ASCII table, so putting offset on space also changes the character.

Opeyem1a (approved):

PR #56, opened 2022-02-26
Fix print vs println bug

### LINKED ISSUE #46: Patch print vs println regex bug

Opeyem1a:
+ make ending with a carriage return optional

Closes #46


aadiraju (approved):
LGTM!

PR #57, opened 2022-02-26
pre defined classes- mathematical programs-hard-q3

SamamthaH:
Wrote out clause for Area Sector, awaiting double literals to finish giving clearer error messages to student


src/pre_defined_classes/mathematical_program/hard/q3/MainTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public void runMain() {
+        AreaSector.main(new String[]{});
+    }
+
+    @Test
+    public void mathTest() {
+        assertEquals(24.801128670839425, Double.parseDouble(getItemByName("math")), 0.1, "Your program's math calculations are incorrect");
+    }
+}
```
Opeyem1a, line 27: EOF newline (this means you're missing a new line at the end of the file, this is good practice for using git version controlling and just in general tbh)

src/pre_defined_classes/mathematical_program/hard/q3/MainTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+import global.BaseTest;
+import global.variables.Clause;
+import global.variables.clauses.*;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MainTest extends BaseTest {
+    //parsons
```
Opeyem1a, line 11: Nitpick but this should be `// Parsons` so it's exactly consistent

Opeyem1a (changes requested):

src/pre_defined_classes/mathematical_program/hard/q3/MainTest.java
SamamthaH: Do we want an extra empty line at the end of file? Also how does this relate to git version control?

src/pre_defined_classes/mathematical_program/hard/q3/MainTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+import global.BaseTest;
+import global.variables.Clause;
+import global.variables.clauses.*;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MainTest extends BaseTest {
+    //Parsons
```
Opeyem1a, line 11: `// Parsons`

src/pre_defined_classes/mathematical_program/hard/q3/MainTest.java
```diff
@@ -0,0 +1,27 @@
...
+    public Clause[] testSentence() {
+        return new Clause[]{
+                new StringLiteral("The area of the sector is: "),
+                new DoubleLiteral("math")
+        };
+    }
+
+    public void runMain() {
+        AreaSector.main(new String[]{});
```
Opeyem1a, line 20: nitpick but I kinda like the String[0] pattern more

src/pre_defined_classes/mathematical_program/hard/q5/MainTest.java
```diff
@@ -0,0 +1,31 @@
+package pre_defined_classes.mathematical_program.hard.q5;
```
Opeyem1a, line 1: Oh wait why is this file here?

src/pre_defined_classes/mathematical_program/hard/q3/MainTest.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    @Test
+    public void mathTest() {
+        assertEquals(24.801128670839425, Double.parseDouble(getItemByName("math")), 0.1, "Your program's math calculations are incorrect");
```
Opeyem1a, line 25: This error message could be improved a bit. Maybe more like:
"Your calculated sector area is incorrect."

Opeyem1a (changes requested):

src/pre_defined_classes/mathematical_program/hard/q5/MainTest.java
SamamthaH: I have no idea. I think something happened while I was messing around with this file while waiting to be able to use negatives and it pushed it when I flipped over to working on this program.

src/pre_defined_classes/mathematical_program/hard/q3/MainTest.java
SamamthaH: Why? They're only doing the one calculation, the output already says that its the calculated sector area

Hedgemon4 (approved):
Good work

src/pre_defined_classes/mathematical_program/hard/q3/MainTest.java
```diff
@@ -0,0 +1,27 @@
...
+        };
+    }
+
+    public void runMain() {
+        AreaSector.main(new String[0]);
+    }
+
+    @Test
+    public void mathTest() {
+        assertEquals(24.801128670839425, Double.parseDouble(getItemByName("area")), 0.001, "Your calculated sector area is incorrect.");
```
aadiraju, line 25: Do we really need all of this specificity? Wouldn't it make sense to limit it to 3 decimal values after the decimal place?
So this would be `24.801` instead and the delta will account for the variation that is smaller than that.

Opeyem1a (approved):

PR #58, opened 2022-02-26
pre defined classes-mathematical program-hard-q4

SamamthaH:
Wrote out clause for Calc Bill, awaiting double literals to finish giving clearer error messages to student


src/pre_defined_classes/mathematical_program/hard/q4/CalcBillTest.java
```diff
@@ -0,0 +1,26 @@
...
+        };
+    }
+
+    public void runMain() {
+        CalcBill.main(new String[]{});
+    }
+
+    @Test
+    public void tipPercentageTest() {
+        assertEquals("28.0", getItemByName("math"), "Your math is incorrect");
```
Hedgemon4, line 24: This test would never fail, as Clause only accepts 28, so it is redundant if the potential values are not widened

Hedgemon4 (changes requested):

src/pre_defined_classes/mathematical_program/hard/q4/CalcBillTest.java
SamamthaH: It's more about providing better feedback than explicitly needing to check the math. If the math is wrong but I didn't have the test they might think their format is the only thing that's wrong.

SanaBob (approved):

src/pre_defined_classes/mathematical_program/hard/q5/MainTest.java
```diff
@@ -0,0 +1,20 @@
+package pre_defined_classes.mathematical_program.hard.q5;
```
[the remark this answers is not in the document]
SamamthaH: Iiiiiii have absolutely no idea I'm gonna get rid of it

src/pre_defined_classes/mathematical_program/hard/q4/CalcBill.java
```diff
@@ -0,0 +1,23 @@
+package pre_defined_classes.mathematical_program.hard.q4;
+
+/*
+Make a Java program that determines the cost of some take-out sushi, given that you tip at 15% and want to round your
```
Hedgemon4, line 4: Maybe we should add that the cost of the sushi is 23.75 to the question

Hedgemon4: As this variable is just called total, which I think is a bit confusing

src/pre_defined_classes/mathematical_program/hard/q4/MainTest.java
```diff
@@ -0,0 +1,27 @@
...
+        };
+    }
+
+    public void runMain() {
+        CalcBill.main(new String[]{});
+    }
+
+    @Test
+    public void tipPercentageTest() {
+        assertEquals(28.0, Double.parseDouble(getItemByName("billAmount")), 0.1, "Your math is incorrect");
```
Hedgemon4, line 25: Again, could say something more like "calculated bill amount is incorrect", but I think this is my preference

Hedgemon4 (changes requested):
I think the question could be a bit more clear, see my comments

src/pre_defined_classes/mathematical_program/hard/q4/CalcBill.java
SamamthaH: I have made some modifications to the variable names and question description which should make it look clearer. I'd appreciate it if you could look it over to see if it makes more sense now.

src/pre_defined_classes/mathematical_program/hard/q4/MainTest.java
```diff
@@ -0,0 +1,27 @@
...
+        };
+    }
+
+    public void runMain() {
+        CalcBill.main(new String[0]);
+    }
+
+    @Test
+    public void tipPercentageTest() {
+        assertEquals(28.0, Double.parseDouble(getItemByName("billAmount")), 0.1, "Your final bill calculation is incorrect.");
```
Hedgemon4, line 25: I think your delta should be smaller here, I don't see it as a problem for this question, but in practice I think that is too wide a range

src/pre_defined_classes/mathematical_program/hard/q4/CalcBill.java
Hedgemon4: I think this is much better now

Hedgemon4: Did you update the document with this as well?

Hedgemon4 (approved):
Looks good

aadiraju (approved):
Looks good to me!

Opeyem1a (approved):

Opeyem1a (approved):

PR #59, opened 2022-02-26
pre defined classes-mathematical program-hard-q5

SamamthaH:


src/pre_defined_classes/mathematical_program/hard/q5/LeftRightTest.java
```diff
@@ -0,0 +1,19 @@
+package pre_defined_classes.mathematical_program.hard.q5;
+
+import global.BaseTest;
+import global.variables.Clause;
+import global.variables.clauses.*;
+
+public class LeftRightTest extends BaseTest {
+    public Clause[] testSentence() {
+        return new Clause[]{
+                new StringLiteral("The leftHandSide is -0.7760683270883323"),
```
Hedgemon4, line 10: Should use DoubleLiteral here so we can check the math

src/pre_defined_classes/mathematical_program/hard/q5/LeftRightTest.java
```diff
@@ -0,0 +1,19 @@
...
+                new NewLine(),
+                new StringLiteral("The rightHandSide is -0.7760683270883322")
```
Hedgemon4, line 12: Same here

Hedgemon4 (changes requested):

src/pre_defined_classes/mathematical_program/hard/q5/LeftRightTest.java
SamamthaH: Looking at the distractors they're all compiler issues, so they cannot possibly get the math wrong, so I didn't see the point in checking that.

src/pre_defined_classes/mathematical_program/hard/q5/LeftRight.java
```diff
@@ -0,0 +1,26 @@
...
+sin( A + B ) = sin( A )cos( B ) + cos( A )sin( B ) where A and B are angles in radians.
+
+Distractors:
+int A = 2.56;
+int B = 1.4312;
+float leftHandSide = Math.sinh( A+B );
+double leftHandSide = sin( A + B );
+double rightHandSide = sin( A ) * cos( B ) + cos( A ) * sin( B );
+double rightHandSide = Math.sin( A )Math.cos( B ) + Math.cos( B )Math.sin( B );
+ */
```
Hedgemon4, line 15: I think you should add a sample output so they know to print left hand side first

src/pre_defined_classes/mathematical_program/hard/q5/MainTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public void runMain() {
+        LeftRight.main(new String[]{});
+    }
+
+    @Test
+    public void mathTest(){
+        assertEquals(-0.7760683270883323, Double.parseDouble(getItemByName("leftSide")), 0.1, "Your calculations for the left hand side is incorrect");
+        assertEquals(-0.7760683270883322, Double.parseDouble(getItemByName("rightSide")), 0.1, "Your calculations for the right hand side is incorrect");
+    }
```
Hedgemon4, line 30: You may want to consider seeing if they equal each other, but that might be overkill

Hedgemon4 (changes requested):

src/pre_defined_classes/mathematical_program/hard/q5/MainTest.java
SamamthaH: Why would I do that?

src/pre_defined_classes/mathematical_program/hard/q5/LeftRight.java
SamamthaH: My clause doesn't check for a specific double output, just a double, and using the delta at all basically means it registers both answers as right for both asserts, so it doesn't really matter

Hedgemon4: No I meant the string for this one, you have the lefthand side one first, and if they switch the order around it will not work

Hedgemon4 (changes requested):

Hedgemon4 (approved):
Good job

src/pre_defined_classes/mathematical_program/hard/q5/MainTest.java
```diff
@@ -0,0 +1,31 @@
...
+        };
+    }
+
+    public void runMain() {
+        LeftRight.main(new String[0]);
+    }
+
+    @Test
+    public void mathTest(){
+        assertEquals(-0.7760683270883323, Double.parseDouble(getItemByName("leftSide")), 0.000001, "Your calculations for the left hand side is incorrect.");
```
Opeyem1a, line 28: `"...are incorrect"` 
For both error messages

Opeyem1a (changes requested):

Opeyem1a (approved):

PR #80, opened 2022-03-04
conditional_simple programs with decision points_hard_q1

SamamthaH:


src/conditionals/simple_programs_with_decision_points/hard/q1/AbsolutelySureTest.java
```diff
@@ -0,0 +1,32 @@
...
+    public Clause[] testSentence() {
+        return new Clause[]{
+                new StringLiteral("Enter a number: "),
+                new NewLine(),
+                new IntegerLiteral(5, "returnedValue")
+        };
+    }
+
+    public void runMain() {
+        provideInput("5");
```
SanaBob, line 20: put 5 in it's own variable at the top of the class and input it here

src/conditionals/simple_programs_with_decision_points/hard/q1/AbsolutelySureTest.java
```diff
@@ -0,0 +1,32 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class AbsolutelySureTest extends BaseTest {
+    public Clause[] testSentence() {
+        return new Clause[]{
+                new StringLiteral("Enter a number: "),
+                new NewLine(),
+                new IntegerLiteral(5, "returnedValue")
```
SanaBob, line 15: same here you can assign 5 to its own variable and input it here which makes it less hardcoded

src/conditionals/simple_programs_with_decision_points/hard/q1/AbsolutelySureTest.java
```diff
@@ -0,0 +1,34 @@
...
+        provideInput(String.valueOf(input));
+        AbsolutelySure.main(new String[]{});
+    }
+
+    @Test
+    public void negativeValueTest() {
+        provideInput("-46");
+        AbsolutelySure.main(new String[]{});
+        String output = getOutput();
+        int contains = output.indexOf("-46");
```
SanaBob, line 31: you can put -46 in its own variable and then provideInpit(variable) and indexOf(variable) to look less hardcoded

src/conditionals/simple_programs_with_decision_points/hard/q1/MainTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import global.BaseTest;
+import global.variables.Clause;
+import global.variables.clauses.*;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class MainTest extends BaseTest {
+    //Parsons
```
Hedgemon4, line 11: `\\ Parsons with distractors`

src/conditionals/simple_programs_with_decision_points/hard/q1/MainTest.java
```diff
@@ -0,0 +1,36 @@
...
+                new IntegerLiteral(input, "returnedValue")
+        };
+    }
+
+    public void runMain() {
+        provideInput(String.valueOf(input));
+        AbsolutelySure.main(new String[]{});
+    }
+
+    @Test
```
Hedgemon4, line 28: You might want to consider using the inputTest beta for this question

Hedgemon4 (changes requested):

src/conditionals/simple_programs_with_decision_points/hard/q1/MainTest.java
```diff
@@ -0,0 +1,46 @@
...
+    static Stream<Arguments> inputProvider() {
+        return Stream.of(Arguments.of(5,5), Arguments.of(-46, 46), Arguments.of(0, 0));
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void testWithInput(int input, int output) {
+        String error = "The absolute value was calculated incorrectly. ";
+        if(input<0 )
+            error += "Negative values were no handled properly.";
```
Hedgemon4, line 41: I presume this should say not instead of no

Hedgemon4: I also feel that I don't think the message is fantastic if you deliver both, maybe you should have different messages depending on what happens

Hedgemon4 (changes requested):

src/conditionals/simple_programs_with_decision_points/hard/q1/MainTest.java
SamamthaH: what do you mean? If it fails a test for a positive value it just says the base error message but then it adds the customized negative one if it doesn't convert the negative to a positive.

src/conditionals/simple_programs_with_decision_points/hard/q1/MainTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+    static Stream<Arguments> inputProvider() {
+        return Stream.of(Arguments.of(5,5), Arguments.of(-46, 46), Arguments.of(0, 0));
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void testWithInput(int input, int output) {
+        String error = "The absolute value was calculated incorrectly. ";
+        if(input<0 )
```
Opeyem1a, line 40: format file

Opeyem1a: If you're going to have custom error messages based on if it handles different cases well, hardcode that as an extra argument from the `inputProvider`. Whether or not you do that, they should be more descriptive if you're going through the trouble of making them distinct. 

i.e. "The absolute value of positive numbers is ..."

Opeyem1a (changes requested):

Opeyem1a (approved):

PR #83, opened 2022-03-06
Input based testing

### LINKED ISSUE #15: Sleeker handling of user input
Idea 1:

1. Modify `BaseTest` to have setOptions(), this is a method that allows users to set values of constants defined in `BaseTest`
i.e.
```java
abstract class BaseTest {
    static boolean autoRunMain = true;
    
    void setOptions() {
        // maybe set default constant values here
        // but default implementation just leaves defaults as is
    }  // this method is called in the @BeforeAll
    ...
}
```

```java
class Question1 {
    void setOptions() {
        Question1.autoRunMain = false;  // or maybe it might be BaseTest.autoRunMain = false;
    }
}
```

We then go to the method in `BaseTest` that runs before each test and have it only executeMain() if the autoRunMain value is true. Otherwise, it doesn't. This lets users trigger main on their own so they can provide input before

- Note: we have to give dev feedback properly when they don't trigger main, like don't let getOutput() do what it does unless main has been run

### LINKED ISSUE #30: For questions with Scanner inputs, we cannot currently run each test with a different input
Since runMain is called once for all of the tests. we can't check differing outputs unless we manually call the methods in order

Opeyem1a:
+ added PlaceHolder clauses & clause injection
+ runWithInput helper method

Closes #30
Closes #15

### DIFF
--- a/src/global/BaseRandomTest.java
+++ b/src/global/BaseRandomTest.java
@@ -1,7 +1,7 @@
 package global;
 
 import global.variables.Clause;
-import global.variables.clauses.RandomClause;
+import global.variables.RandomClause;
 import org.junit.jupiter.api.*;
 
 import java.util.regex.Matcher;
--- a/src/global/BaseTest.java
+++ b/src/global/BaseTest.java
@@ -1,12 +1,9 @@
 package global;
 
-import static global.tools.CustomAssertions._assertTrue;
-import static global.tools.CustomAssertions._fail;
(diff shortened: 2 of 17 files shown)


aadiraju:
changed this PR to merge into a beta branch for testing before we actually merge into main

aadiraju (approved):
I think this structure makes sense. We should put it into Beta first just to be safe though

PR #86, opened 2022-03-11
pre_defined_classes-mathematical_program-medium-q7

Hedgemon4:


SamamthaH (approved):

src/pre_defined_classes/mathematical_program/medium/q7/MainTest.java
```diff
@@ -0,0 +1,28 @@
...
+import global.variables.clauses.StringLiteral;
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MainTest extends BaseTest {
+    // Parsons
+    public Clause[] testSentence() {
+        return new Clause[]{
+                new StringLiteral("The speed of a tsunami starting at a depth of 5000 is "),
+                new DoubleLiteral(220, 222, "waterSpeed"),
```
Opeyem1a, line 15: I think if you want to use DoubleLiteral to grab a specific value, then we shouldn't provide a range like this, I think it's best here to just not provide a range and let it catch any double.

My reasoning is that the default error message for when a number is outside the range will be something like "value for X is outside of range", but the student will have no idea what range we're talking about and the question has no ranged component. Best to reserve the range to cases where the range is critical

Opeyem1a (changes requested):

src/pre_defined_classes/mathematical_program/medium/q7/MainTest.java
```diff
@@ -0,0 +1,28 @@
...
+        };
+    }
+
+    public void runMain() {
+        FastAndWatery.main(new String[0]);
+    }
+
+    @Test
+    public void waterSpeedMathTest(){
+        assertEquals(221.4 , Double.parseDouble(getItemByName("waterSpeed")), 0.2,  "Your calculated water speed is incorrect");
```
Opeyem1a, line 26: Format file and period at the end of the sentence

Opeyem1a (changes requested):

aadiraju (approved):
This is good now!

Opeyem1a (approved):

Opeyem1a (approved):

PR #88, opened 2022-03-11
pre_defined_classes-mathematical-program-medium-q8

Hedgemon4:


src/pre_defined_classes/mathematical_program/medium/q8/MainTest.java
```diff
@@ -0,0 +1,30 @@
...
+import global.variables.clauses.StringLiteral;
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MainTest extends BaseTest {
+    //Parsons
+    public Clause[] testSentence() {
+        return new Clause[]{
+                new StringLiteral("The magnitude of an earthquake with intensity "),
+                new DoubleLiteral(23200.0, 23200.0, "intensity"),
```
SamamthaH, line 15: Is there a possibly different intensity line? If not you could just scrap the two 23200.0s. Either way, I'd say get rid of the name since you never use it

Hedgemon4: No, but I just name everything for consistency

Hedgemon4: Is this ok @Opeyem1a ?

Opeyem1a: I think naming things like this is okay since it makes the code more readable, but it's not necessary for sure. I think when the double is a specific double though we should just catch any double (so no need to set a range)

src/pre_defined_classes/mathematical_program/medium/q8/MainTest.java
```diff
@@ -0,0 +1,30 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MainTest extends BaseTest {
+    // Parsons
+    public Clause[] testSentence() {
+        return new Clause[]{
+                new StringLiteral("The magnitude of an earthquake with intensity "),
+                new DoubleLiteral(23200.0, 23200.0, "intensity"),
+                new StringLiteral(" is "),
+                new DoubleLiteral( "magnitude"),
```
Opeyem1a, line 17: Format this file with `CTRL`+`ALT`+`L` to remove weird spaces like this

Opeyem1a (changes requested):

src/pre_defined_classes/mathematical_program/medium/q8/Earthquake.java
```diff
@@ -0,0 +1,17 @@
...
+earthquakes based on their intensity. The formula to determine the magnitude of an earthquake is: M = log(II0), where I
+is the intensity of that earthquake and I0 is the baseline intensity of an earthquake with magnitude 0
+(and hence always has a value of 1). Make a Java program that prints out the magnitude of an earthquake that had an
+intensity of 2.32*10^4 units.
+ */
+
+public class Earthquake {
+    public static void main(String[] args) {
+        double intensity = 23200;
+        double magnitude = Math.log(intensity / 1);
```
SamamthaH, line 14: Why is this /1 when the question says the formula is log(II0)? Not a you problem, a question problem

SamamthaH (commented):
The test is good, just one issue with the question I think

SamamthaH (approved):

PR #97, opened 2022-03-24
Conditionals simple programs with decision points hard q3

SamamthaH:


SamamthaH:
Bad question is here, since the distractors need to be updated to reflect the changed answer if statement

aadiraju:
I think your tests work as they should so it looks good. I think maybe instead of calling `getOutput()` and checking whether that contains the optional field, we should have a way to deal with this type of case. This works for now though, and we might try to find a better solution for this and implement it

src/conditionals/simple_programs_with_decision_points/hard/q3/MainTest.java
```diff
@@ -0,0 +1,69 @@
...
+    static Stream<Arguments> inputProviderEarly() {
+        return Stream.of(Arguments.of(20, 56), Arguments.of(18, 1));
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProviderLate")
+    void testWithInputLate(int hours, int minutes) {
+        runWithInput((hours + "") + " " + (minutes + ""));
+        String output = getOutput();
+        assertTrue(output.contains("Please go to sleep."), "Your program tells the user to stay up when they should be going to sleep.");
```
Hedgemon4, line 52: Could you not just name the stringLiteral, and then get the one you want from the output as you already split it up into two categories. This would also work for one of the earlier pull requests you had that I said we should talk about as it was returning true/false

src/conditionals/simple_programs_with_decision_points/hard/q3/SleepyTime.java
```diff
@@ -0,0 +1,35 @@
+package conditionals.simple_programs_with_decision_points.hard.q3;
+
+import java.util.Scanner;
+
+/*
+Write a program to ask for the current time, in hours and minutes, and tell the user to go to sleep if it is past 21:30.
+(Assume the use of a 24h clock)
+
```
Hedgemon4, line 8: I think there should be a sample output here

Hedgemon4 (changes requested):

src/conditionals/simple_programs_with_decision_points/hard/q3/MainTest.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    static Stream<Arguments> inputProvider() {
+        return Stream.of(Arguments.of(22, 31, "Please go to sleep."), Arguments.of(22, 24, "Please go to sleep."), Arguments.of(20, 56, "The night is still young!"), Arguments.of(18, 1, "The night is still young!"));
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void testWithInput(int hours, int minutes, String message) throws InvalidClauseException {
+        runWithInput((hours + " ") + (minutes + ""), new Clause[]{
```
Opeyem1a, line 41: `hours + " " + minutes` should be fine, no?

Opeyem1a (changes requested):

Opeyem1a (approved):

PR #98, opened 2022-03-25
conditionals-simple_programs_with_decision_points-hard-q4

SamamthaH:


Hedgemon4 (approved):
Looks good

src/conditionals/simple_programs_with_decision_points/hard/q4/MainTest.java
```diff
@@ -0,0 +1,44 @@
...
+                new PlaceHolder()
+        };
+    }
+
+    public void runMain() {
+        DoorBoy.main(new String[0]);
+    }
+
+    static Stream<Arguments> inputProvider() {
+        return Stream.of(Arguments.of("true", "I will close the door"), Arguments.of("false", "Time to open the door"));
```
Opeyem1a, line 34: Using input provider with only 2 cases seems like a waste, I think it'd be better with just 2 tests, one for each case

I think 2 separate tests with try-catch around the runWithInput is the better move so we can give specific feedback to the student for their mistake. (dw if this doesn't make sense, ask Abhi or I when you get to it)

Opeyem1a (changes requested):

src/conditionals/simple_programs_with_decision_points/hard/q4/MainTest.java
```diff
@@ -0,0 +1,42 @@
...
+        };
+    }
+
+    public void runMain() {
+        DoorBoy.main(new String[0]);
+    }
+
+    @Test
+    public void testOpen() throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program should have closed the door, it was already open.";
```
Opeyem1a, line 29: The error messages in this file would be weird to read from a student's perspective since they won't (once we fix it) know what the input was that caused the test to fail.

Perhaps something more like:
"Your program attempted to (open/close) an already (opened/closed) door."

Opeyem1a (changes requested):
Other than this, solid

PR #105, opened 2022-03-25
Fix negative integer issue

### LINKED ISSUE #104: negative integers don't work as intended

Opeyem1a:
+ allow IntegerLiteral to always match negative numbers

> We'll need to additionally merge this into `beta` to ensure the change persist there too

Closes #104


aadiraju (approved):
This is a good fix for this issue. Good job

PR #113, opened 2022-04-01
 conditionals-simple_programs_with_decision_points-medium-q7

Hedgemon4:


SamamthaH (approved):

SamamthaH (approved):

Opeyem1a (approved):

src/conditionals/simple_programs_with_decision_points/medium/q7/MainTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of(100.0, "Remaining HP\\: 100\\.0 \\/ 100"), Arguments.of(0, "Not Alive"),
+                Arguments.of(69.42, "Remaining HP\\: 69\\.42 \\/ 100"), Arguments.of(-5.0, "Not Alive"));
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void correctHPMessage(double hpInput, String output) throws InvalidClauseException {
+        runWithInput(hpInput + "", new Clause[]{
```
Opeyem1a, line 40: Check what the actual error message is for this, is it descriptive? If it just says the generic incorrect structure message then lmk

Opeyem1a (changes requested):

src/conditionals/simple_programs_with_decision_points/medium/q7/MainTest.java
Hedgemon4: It just gives the generic error message

src/conditionals/simple_programs_with_decision_points/medium/q7/MainTest.java
```diff
@@ -0,0 +1,49 @@
...
+                Arguments.of(69.42, "Remaining HP\\: 69\\.42 \\/ 100"), Arguments.of(-5.0, "Not Alive"));
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void correctHPMessage(double hpInput, String output) throws InvalidClauseException {
+        String errorMessage = "Ensure that if HP is less than or equal to zero the message \"Not Alive\" is printed";
+        if(hpInput > 0)
+            errorMessage = "Ensure that the correct remaining HP is printed";
+        TestOption.incorrectStructureErrorMessage =  errorMessage;
```
SamamthaH, line 43: Why do this one instead of the try-catch? Is there a way the altered error message could trigger for what is actually a structure error? In which case things would get even more confusing.

Opeyem1a: This is the optimal pattern now, I mentioned the try catch thing but it's more complex than this, which solves the same problem (I wrote the try catch comments before developing this pattern)

Opeyem1a (approved):

PR #120, opened 2022-04-11
Custom error messages when running with input

Opeyem1a:
+ manage the default structure error message in the test option class
+ reset to the default message before each test
+ allow test writers to define a custom error message for a test

Closes #120


aadiraju (approved):
Looks good!

PR #171, opened 2022-05-31
Improve random testing

### LINKED ISSUE #133: Fix randomness checks failing
Invalid Random Values at 5

org.opentest4j.AssertionFailedError: You do not seem to be properly generating random values. ==> 
Expected :true
Actual   :false

Opeyem1a:
+ switched to use simpler random checking (old random checking methods not yet removed though, located in RandomUtil)
+ added a print of all values that caused the random check to fail for replicability of errors

Closes #133

### DIFF
--- a/src/global/utils/RandomUtil.java
+++ b/src/global/utils/RandomUtil.java
@@ -1,5 +1,6 @@
 package global.utils;
 
+import global.tools.Logger;
 import global.variables.*;
 import org.apache.commons.math3.stat.inference.ChiSquareTest;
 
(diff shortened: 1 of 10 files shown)


src/global/utils/RandomUtil.java
```diff
@@ -64,4 +66,26 @@ public static RandomClause<?> castRandomClause(Clause clause) {
...
         return (RandomClause<?>) clause;
     }
+
+    /**
+     * Determine if a list of frequencies is random
+     *
+     * @param frequencies a list of frequency counts for bins
+     * @return whether or not this array of frequncies describes randomly generated values
+     */
+    public static boolean frequenciesAreRandom(int[] frequencies, int numBins) {
```
aadiraju, line 76: Do we even need `numbins` as a param here? Isn't the frequency array always going to be of size `numBins`?

Opeyem1a: Yes, tbh idk I was wary of having too many implicit assumptions that someone would have to read comments/already understand to comprehend - so I opted to just pass in the value since it's readily available and already calculated

keyvankhademi (approved):

PR #222, opened 2022-06-23
pre_defined_classes-simple_programs_taking_numeric_user_input-medium-q7 

yuqi88:
## PR Checklist
- [x]  EOF newlines
    - [x]  What about edge cases? (empty strings, negatives, infinity, etc)
---

## Not yet, but soon

### DIFF
--- a/src/pre_defined_classes/simple_programs_taking_numeric_user_input/medium/q7/MainTest.java
+++ b/src/pre_defined_classes/simple_programs_taking_numeric_user_input/medium/q7/MainTest.java
@@ -1,13 +1,5 @@
 package pre_defined_classes.simple_programs_taking_numeric_user_input.medium.q7;
 
-
-/*
-Sample output:
-Enter your score:
--5
-Your goal is 45. Keep up the good work!
- */
-
 import global.BaseRandomTest;
 import global.exceptions.InvalidClauseException;
 import global.tools.TestOption;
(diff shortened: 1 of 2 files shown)


src/pre_defined_classes/simple_programs_taking_numeric_user_input/medium/q7/MainTest.java
```diff
@@ -0,0 +1,47 @@
...
+import global.tools.TestOption;
+import global.variables.Clause;
+import global.variables.clauses.*;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.Random;
+import java.util.stream.Stream;
+
+public class MainTest extends BaseTest {
```
Hedgemon4, line 22: This should extend BaseRandomTest so you can use the RandomInteger() clause instead

src/pre_defined_classes/simple_programs_taking_numeric_user_input/medium/q7/MainTest.java
```diff
@@ -0,0 +1,47 @@
...
+    //Parsons
+    public Clause[] testSentence(){
+        TestOption.isInputTest = true;
+        TestOption.defaultInput = "-5";
+        return new Clause[]{
+                new StringLiteral("Enter your score: "),
+                new NewLine(),
+                new StringLiteral("Your goal is "),
+                new IntegerLiteral("goal"),
```
Hedgemon4, line 31: This is where you can use RandomInteger, which tells the framework to check if this number is randomly distributed

src/pre_defined_classes/simple_programs_taking_numeric_user_input/medium/q7/MainTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public void runMain(){ AbsoluteNumber.main(new String[0]); }
+
+    public static Stream<Integer> inputProvider(){
+        return Stream.of( 3 , 0 , -5 );
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    public void correctlyPrintsRandomAbsValue(int input) throws InvalidClauseException{
```
Hedgemon4, line 44: Good method name

Hedgemon4 (changes requested):

src/pre_defined_classes/simple_programs_taking_numeric_user_input/medium/q7/MainTest.java
```diff
@@ -0,0 +1,47 @@
...
+import global.variables.Clause;
+import global.variables.clauses.*;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.Random;
+import java.util.stream.Stream;
+
+public class MainTest extends BaseTest {
+    //Parsons
```
Hedgemon4, line 23: Also this is very minor, but there should be a space between the // and Parsons (`// Parsons`)

src/pre_defined_classes/simple_programs_taking_numeric_user_input/medium/q7/MainTest.java
yuqi88: If I used RandomInteger(), then we will encounter the problem of having to define the upper bound of the random number. To resolve that issue I changed the question so that the input has a range.

yuqi88:
@Hedgemon4 I talked to Opey about this question. We have to change the framework to test this question. 
Because currently, RandomInteger( ) tests if there are numbers in each bin ( a range) equally.
However, in our question, it's testing the output =input * random number. In the case of input = -10, the outputs are always going to ten times a random number between 1 - 10 (inclusive), so not all bin gets filled, and so the test never passes.
problem of num_input_med q7

Hedgemon4:
Waiting for https://github.com/canvas-gamification/canvas-gamification-junit-tests/issues/299

Hedgemon4:
Changed question to output random numbers to avoid "softRandom" option

Hedgemon4 (approved):

PR #223, opened 2022-06-24
Added additional functionality to MethodUtil

Hedgemon4:
… easier, and added support for custom fail messages for invalid methods

## PR Checklist
- [ ]  EOF newlines
    - [ ]  What about edge cases? (empty strings, negatives, infinity, etc)
---

## Not yet, but soon

### DIFF
--- a/src/global/utils/MethodUtil.java
+++ b/src/global/utils/MethodUtil.java
@@ -28,17 +28,7 @@ public class MethodUtil {
     }
 
     public static Object invokeIfMethodExists(Class<?> methodClass, String methodName) {
-        setUpMethodOutput();
-        try {
-            Method testMethodInvoke = methodClass.getMethod(methodName);
-            return testMethodInvoke.invoke(null);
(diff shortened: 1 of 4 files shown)


src/global/utils/MethodUtil.java
```diff
@@ -1,32 +1,52 @@
...
+        } catch (NoSuchMethodException | InvocationTargetException | IllegalAccessException e) {
+            fail(Objects.requireNonNullElseGet(
+                    TestOption.invalidMethodMessage,
+                    () -> String.join("", methodClass.getSimpleName(), " does not contain method ", methodName, "."))
+            );
+            return null;
+        }
+    }
+
+    public static Object invokeIfMethodExists(Class<?> methodClass, String methodName) {
```
Opeyem1a, line 30: I think this method could just call the other one with nulls for the last 2 parameters

Opeyem1a (changes requested):

src/global/utils/MethodUtil.java
Hedgemon4: Makes sense, I will make that change

PR #236, opened 2022-06-30
pre_defined_classes-simple_programs_taking_numeric_user_input-medium-q9

yuqi88:
## PR Checklist
- [ ]  EOF newlines
    - [ ]  What about edge cases? (empty strings, negatives, infinity, etc)
---

## Not yet, but soon


yuqi88:
@Hedgemon4 The question asks for 3 digits and then prints out the reverse version. 
The current solution code takes `.nextInt()`, but in the case of `input = 120`, 
the reverse will be printed as `012`. Because it's typed integer. I don't think this is right.

Should we change the solution code to `.nextLine()` and reverse the input as a string instead?

Hedgemon4:
> @Hedgemon4 The question asks for 3 digits and then prints out the reverse version. The current solution code takes `.nextInt()`, but in the case of `input = 120`, the reverse will be printed as `012`. Because it's typed integer. I don't think this is right.
> 
> Should we change the solution code to `.nextLine()` and reverse the input as a string instead?

I think that because this is in the numeric user input section, it should still take in an int, but then outputs a String instead of an int. We can just change it so it is `String reversePlate = "" + num3 + num2 + num1` and then print that.

src/pre_defined_classes/simple_programs_taking_numeric_user_input/medium/q9/ReversePlate.java
```diff
@@ -0,0 +1,30 @@
+package pre_defined_classes.simple_programs_taking_numeric_user_input.medium.q9;
+
+/*
+Traffic speed cameras are used to catch drivers who speed. One of the issues with this system is that it reads the plate
```
Hedgemon4, line 4: Also, I don't imagine this will work with negative numbers, so we should probably specify that the numbers will be positive.

Hedgemon4 (changes requested):

src/pre_defined_classes/simple_programs_taking_numeric_user_input/medium/q9/MainTest.java
```diff
@@ -0,0 +1,52 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/*
+Sample output:
+    What is the 3 digit plate number?
+    678
+    The reverse number is: 876
+ */
+public class MainTest extends BaseTest {
+    //Parson
```
Hedgemon4, line 25: `// Parsons`

Hedgemon4 (changes requested):
Change parsons comment

src/pre_defined_classes/simple_programs_taking_numeric_user_input/medium/q9/ReversePlate.java
```diff
@@ -0,0 +1,34 @@
...
+ numbers backwards. For simplicity, suppose that all the plates are 3 digits long.
+ Make a program to reverse an inputted plate number.
+ Assuming all user inputs are valid, meaning that all 3 digits are either positive or zero
+ and that all 3 digits are entered ( plate 008 should have the 2 zero enter too. 8 is not a valid plate.)
```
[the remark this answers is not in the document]
Hedgemon4: Also, the way this is now, would it not print 008? I think this is a valid plate

src/pre_defined_classes/simple_programs_taking_numeric_user_input/medium/q9/MainTest.java
```diff
@@ -0,0 +1,54 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/*
+Sample output:
+    What is the 3 digit plate number?
+    678
+    The reverse number is: 876
+ */
+public class MainTest extends BaseTest {
+    // Parson
```
Hedgemon4, line 25: `// Parsons`

Hedgemon4 (changes requested):

src/pre_defined_classes/simple_programs_taking_numeric_user_input/medium/q9/ReversePlate.java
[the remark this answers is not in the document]
yuqi88: if the user typed 8, the program will treat it as 008, so the output will be 800 which makes sense from a programming perspective. But from a user's perspective, it's weird. It's like they typed in an unfinished input, accidentally press enter and still got a valid output.

yuqi88:
There's a issue of using `String reversePlate = "" + num3 + num2 + num1`.
It works for input = 800. We will get a correct output = 008. 
But it does not work for input = 008. We will get a incorrect output = 80000.

Using `String reversePlate = num3 + num2 + num1 + ""` will produce a similar result except input = 800 becomes incorrect & input 008 becomes correct this time.

To solve this issue, I'll add an assumption to the question that no zeros are allowed on the plates.

src/pre_defined_classes/simple_programs_taking_numeric_user_input/medium/q9/ReversePlate.java
```diff
@@ -0,0 +1,33 @@
+package pre_defined_classes.simple_programs_taking_numeric_user_input.medium.q9;
+
+/*
+Traffic speed cameras are used to catch drivers who speed. One of the issues with this system is that it reads the plate
+numbers backwards. For simplicity, suppose that all the plates are 3 digits long.
+Make a program to reverse an inputted plate number.
+Assuming all users enter exactly 3 digits from 1 to 9 (Zeros are not allowed).
```
Hedgemon4, line 7:
```suggestion
Assume all users enter exactly 3 digits from 1 to 9 (Zeros are not allowed).
```

src/pre_defined_classes/simple_programs_taking_numeric_user_input/medium/q9/MainTest.java
```diff
@@ -0,0 +1,44 @@
...
+import global.variables.clauses.PlaceHolder;
+import global.variables.clauses.StringLiteral;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
+
+public class MainTest extends BaseTest {
+    // Parson
```
Hedgemon4, line 17:
```suggestion
    // Parsons
```

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #271, opened 2022-07-09
pre_defined_classes-simple_programs_taking_string_user_input-hard-q9

yuqi88:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/pre_defined_classes/simple_programs_taking_string_user_input/hard/q9/Origin.java
```diff
@@ -0,0 +1,43 @@
+package pre_defined_classes.simple_programs_taking_string_user_input.hard.q9;
+
+/*
+People all around the world use last names to identify their origins. Common last names include:"Silva" from Portugal
```
Hedgemon4, line 4: This is just a very strange question, I am going to ask Bowen if we can just get rid of this one.

Hedgemon4 (approved):

PR #274, opened 2022-07-11
pre_defined_classes-programs_with_changing_behaviors-medium-q7

yuqi88:
## PR Checklist
- [ ]  EOF newlines
---
## Website


Hedgemon4 (approved):
Looks good

PR #279, opened 2022-07-11
conditionals-programs_with_decision_points-hard-q8

ketphan02:
## PR Checklist
- [ ]  EOF newlines
---
## Website

### DIFF
--- a/src/conditionals/programs_with_decision_points/hard/q8/MainTest.java
+++ b/src/conditionals/programs_with_decision_points/hard/q8/MainTest.java
@@ -17,7 +17,7 @@ public class MainTest extends BaseTest {
   // Parsons with distractors
   public Clause[] testSentence() {
     TestOption.isInputTest = true;
-    TestOption.defaultInput = "1 1";
+    TestOption.defaultInput = "500";
     return new Clause[] {
         new StringLiteral("How many points did you score\\?"),
         new NewLine(),
--- a/src/conditionals/programs_with_decision_points/hard/q8/MainTest.java
+++ b/src/conditionals/programs_with_decision_points/hard/q8/MainTest.java
@@ -21,7 +21,6 @@ public class MainTest extends BaseTest {
     return new Clause[] {
         new StringLiteral("How many points did you score\\?"),
         new NewLine(),
-        new PlaceHolder(),
         new PlaceHolder()
     };
   }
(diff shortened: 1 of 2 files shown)


src/conditionals/programs_with_decision_points/hard/q8/PointsScoring.java
```diff
@@ -0,0 +1,44 @@
+package conditionals.programs_with_decision_points.hard.q8;
+
+/*
+ * Write a Java Program that asks a player for their score, and prints out the prize that a player wins in a game
+ * depending on their score. If the player scores 5000 points or above, they win a gold prize. If the player scores
+ * between 2500 - 4999 points, they win a silver prize. If the player scores between 1000 - 2499 points, they win a
+ * bronze prize. If they score below 1000 points, they don't win any prize. Make sure to print out an error message if 
+ * the player enters an invalid score.
```
Hedgemon4, line 8: I would change the `Make sure to print an error message ... ` sentence because the question does not specify what is a valid score. If we want to make sure the score is >=  0, just change it to `Make sure to check if the score is greater than or equal to 0, and print out an error message if it is not.`

src/conditionals/programs_with_decision_points/hard/q8/MainTest.java
```diff
@@ -0,0 +1,72 @@
...
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
+
+public class MainTest extends BaseTest {
+  // Parsons with distractors
+  public Clause[] testSentence() {
+    TestOption.isInputTest = true;
+    TestOption.defaultInput = "1 1";
```
Hedgemon4, line 20: Why is this the default input?

Hedgemon4 (changes requested):

src/conditionals/programs_with_decision_points/hard/q8/MainTest.java
ketphan02: My bad :(

src/conditionals/programs_with_decision_points/hard/q8/MainTest.java
```diff
@@ -0,0 +1,72 @@
...
+    TestOption.incorrectStructureErrorMessage = "Your program does not print the correct prize for the given score.";
+    runWithInput(score + " ", new Clause[] {
+        new StringLiteral(prize)
+    });
+  }
+
+  @ParameterizedTest
+  @MethodSource("inputProviderInvalid")
+  public void printsErrorMessageWithInvalidInput(int score) throws InvalidClauseException {
+    TestOption.incorrectStructureErrorMessage = "Your program does not print an error message when given invalid score.";
```
Hedgemon4, line 67: Add an to error message (`... when given an invalid score.`)

Hedgemon4 (changes requested):

Hedgemon4 (approved):
Good work

PR #286, opened 2022-07-12
 pre_defined_classes-programs_with_changing_behaviors-hard-q8

yuqi88:
## PR Checklist
- [x]  EOF newlines
---
## Website


yuqi88:
Issue related to question design:
1. A starting position at 46 with 2 dices will only range from 48 to 58 for the next roll. Should we delete the description of where all the other snakes are? It's unnecessary info in the question.
2. According to the current question(with starting position at 46), if the two dice add up to 8 (the player gets 54). The actual position the player will end up in should be 54-19 = 35 because of the snake on 54. Seems like this question will need an if-statement or we need to change where the question to another board game. Thoughts?

yuqi88:
Issue related to RandomInteger( ):
With 2 dices, the user can have +2 to +12 steps. The chances of getting the middle range are larger than the ends. Though the 2 dices are randomly generated with the values from 1 to 6, the added values are not equally distributed; thus causing the test not to pass.

Explanation with an example:
The chance of getting + 7 can result from 1+6, 2+5, 3+4, 4+3, 5+2, or 6+1. Whereas +2 can only result from 1+1.

Hedgemon4:
> 

Right. I think this could also be solved by something we are talking about implementing, but I am not sure yet

src/pre_defined_classes/programs_with_changing_behaviour/hard/q8/SnakesAndLadders.java
```diff
@@ -0,0 +1,34 @@
+package pre_defined_classes.programs_with_changing_behaviour.hard.q8;
+
```
Hedgemon4, line 2: I think for now, we should just leave this one and I will come back to it.

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #289, opened 2022-07-13
loops-while_loops-hard-q6

SamamthaH:
## PR Checklist
- [ ]  EOF newlines
---
## Website


src/loops/while_loops/hard/q6/MainTest.java
```diff
@@ -0,0 +1,42 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.stream.Stream;
+
+public class MainTest extends BaseTest {
+    //Parsons
```
Hedgemon4, line 16: `// Parsons`

src/loops/while_loops/hard/q6/FactOrIal.java
```diff
@@ -0,0 +1,25 @@
+package loops.while_loops.hard.q6;
+
+/*
+Write a program that will accept a non-negative number from the user and print out the factorial of that number.
```
Hedgemon4, line 4: As this is after conditionals, we can actually just make them check if the number is >= 0.

Hedgemon4: So basically we can have 
```
if(input < 0)
  System.out.println("Invalid input!)
```

src/loops/while_loops/hard/q6/FactOrIal.java
```diff
@@ -0,0 +1,25 @@
...
+Do so by using a while loop. ( Hint: n factorial or n! = n x (n - 1) x (n - 2) x … x 2 x 1 )
+ */
+import java.util.Scanner;
+public class FactOrIal
```
Hedgemon4, line 8: I'm not a fan of the random upper case class name

src/loops/while_loops/hard/q6/MainTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of(0, 0), Arguments.of(10, 3628800), Arguments.of(5, 120), Arguments.of(1, 1), Arguments.of(12, 479001600));
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void calculatesFactorialCorrectly(int in, int factorial){
+        runWithInput(in + System.lineSeparator());
+        assertEquals(Integer.parseInt(getItemByName("number")), in, "Your program does not output the same number that was input");
```
Hedgemon4, line 39: I know what you are trying to say with the error message, but I think it needs to be tweaked a bit

src/loops/while_loops/hard/q6/MainTest.java
```diff
@@ -0,0 +1,42 @@
...
+        assertEquals(Integer.parseInt(getItemByName("factorial")), factorial, "Your program does not correctly calculate the factorial");
```
Hedgemon4, line 40: This one is good, but is missing a period

src/loops/while_loops/hard/q6/MainTest.java
```diff
@@ -0,0 +1,42 @@
...
+                new IntegerLiteral("factorial")
+        };
+    }
+
+    public void runMain() { FactOrIal.main(new String[0]); }
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of(0, 0), Arguments.of(10, 3628800), Arguments.of(5, 120), Arguments.of(1, 1), Arguments.of(12, 479001600));
+    }
+
```
Hedgemon4, line 34: Once you add the input check, I would add another test for invalid inputs

Hedgemon4 (changes requested):

src/loops/while_loops/hard/q6/FactOrIal.java
SamamthaH: I am not the one that picked the class name, that's just what was in the document. Do you want me to change the document?

src/loops/while_loops/hard/q6/MainTest.java
```diff
@@ -0,0 +1,54 @@
...
+                new PlaceHolder()
+        };
+    }
+
+    public void runMain() { Factorial.main(new String[0]); }
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of(0, 0, 3), Arguments.of(10, 3628800, 3), Arguments.of(5, 120, 3), Arguments.of(1, 1, 3), Arguments.of(12, 479001600, 3), Arguments.of(-6, -1, 1));
+    }
+
```
Hedgemon4, line 35: I think you should have a separate test called something like `printsErrorMessageForInvalidInput()` so that in the end, the student would get the output of 
```
What went well:
- Calculates Factorial Correctly
- Prints Error Message ForI nvalidI nput
```

Hedgemon4 (changes requested):

src/loops/while_loops/hard/q6/MainTest.java
```diff
@@ -0,0 +1,56 @@
...
+        place[0][2] = new IntegerLiteral("factorial");
+        runWithInput(in + System.lineSeparator(), place);
+        assertEquals(Integer.parseInt(getItemByName("number")), in, "Your program does not output the input number in the final output.");
+        assertEquals(Integer.parseInt(getItemByName("factorial")), factorial, "Your program does not correctly calculate the factorial.");
+    }
+
+    @Test
+    void printsErrorMessageForInvalidInput() throws InvalidClauseException{
+        TestOption.incorrectStructureErrorMessage = "Your program does not identify an invalid input.";
+        runWithInput(-6 + System.lineSeparator(), new Clause[]{
```
Hedgemon4, line 52: This works as it is a `// Parsons` question, but it would be better to check the boundary condition (-1 in this case)

Hedgemon4 (approved):

PR #295, opened 2022-07-15
loops-programs_with_repetition-easy-q4

ketphan02:
## PR Checklist
- [x]  EOF newlines
---
## Website

### DIFF
--- a/src/loops/programs_with_repetition/easy/q4/ItsAlwaysPrimeTime.java
+++ b/src/loops/programs_with_repetition/easy/q4/ItsAlwaysPrimeTime.java
@@ -1,7 +1,7 @@
 package loops.programs_with_repetition.easy.q4;
 
 /*
- * Write a program to keep asking the user to enter a number until they enter a prime number that is lesser than 50. Make sure your output matches the sample. 
+ * Write a program to keep asking the user to enter a number until they enter a prime number that is less than 50. Make sure your output matches the sample. 
  * Note that prime numbers should not be negative. 
  * Sample output:
  *    Enter an integer:
--- a/src/loops/programs_with_repetition/easy/q4/ItsAlwaysPrimeTime.java
+++ b/src/loops/programs_with_repetition/easy/q4/ItsAlwaysPrimeTime.java
@@ -5,6 +5,12 @@ package loops.programs_with_repetition.easy.q4;
  * 
  * Sample output:
  *    Enter an integer:
+ *    1
+ *    Enter an integer:
+ *    6
+ *    Enter an integer:
+ *    50
+ *    Enter an integer:
  *    2
  *    Entered a prime number that is less than 50. Ending Program.
  */
--- a/src/loops/programs_with_repetition/easy/q4/ItsAlwaysPrimeTime.java
+++ b/src/loops/programs_with_repetition/easy/q4/ItsAlwaysPrimeTime.java
@@ -1,7 +1,7 @@
 package loops.programs_with_repetition.easy.q4;
 
 /*
- * Write a program to keep asking the user to enter a number until they enter a prime number that is lesser than 50
+ * Write a program to keep asking the user to enter a number until they enter a prime number that is lesser than 50. Make sure your output matches the sample. 
  * 
  * Sample output:
  *    Enter an integer:
--- a/src/loops/programs_with_repetition/easy/q4/ItsAlwaysPrimeTime.java
+++ b/src/loops/programs_with_repetition/easy/q4/ItsAlwaysPrimeTime.java
@@ -1,36 +1,38 @@
 package loops.programs_with_repetition.easy.q4;
 
 /*
- * Write a program to keep asking the user to enter a number until they enter a prime number that is less than 50. Make sure your output matches the sample. 
- * Note that prime numbers should not be negative. 
- * Sample output:
- *    Enter an integer:
- *    1
- *    Enter an integer:
- *    6
- *    Enter an integer:
- *    50
- *    Enter an integer:
- *    2
- *    Entered a prime number that is less than 50. Ending Program.
+Write a program to keep asking the user to enter a number until they enter a prime number that is less than 50. Once a
+valid input is received, the program should display the message "Entered a prime number that is less than 50. Ending Program.".
+Note that prime numbers should not be negative. Make sure your output matches the sample.
+
+Sample output:
+    Enter an integer:
+    1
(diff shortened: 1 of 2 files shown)


src/loops/programs_with_repetition/easy/q4/MainTest.java
```diff
@@ -0,0 +1,86 @@
...
+    Iterator<Integer> it = inputs.iterator();
+    Clause[][] outputStrings = new Clause[1][inputs.size() * 3 - 2];
+    for (int i = 0; i < inputs.size() * 3 - 3; i += 3) {
+      outputStrings[0][i + 0] = new IntegerLiteral(it.next());
+      outputStrings[0][i + 1] = new StringLiteral("Enter an integer\\: ");
+      outputStrings[0][i + 2] = new NewLine();
+    }
+    outputStrings[0][inputs.size() * 3 - 3] = new StringLiteral(
+        "Entered a prime number that is less than 50. Ending Program. ");
+  }
```
Hedgemon4, line 85: This is missing a runWithInput

src/loops/programs_with_repetition/easy/q4/MainTest.java
```diff
@@ -0,0 +1,86 @@
...
+import global.BaseTest;
+import global.exceptions.InvalidClauseException;
+import global.tools.TestOption;
+import global.variables.Clause;
+import global.variables.clauses.IntegerLiteral;
+import global.variables.clauses.NewLine;
+import global.variables.clauses.PlaceHolder;
+import global.variables.clauses.StringLiteral;
+
+public class MainTest extends BaseTest {
```
Hedgemon4, line 20: I think what you did here with the random input is a good idea; however because you did it randomly, there is a chance that a program which should not pass will if a negative or a number above 50 is not selected.

Hedgemon4: Generally, we want to hardcode some base cases for things like boundaries and general logic, and then we can add a few random cases to catch hardcoding if necessary.

src/loops/programs_with_repetition/easy/q4/MainTest.java
```diff
@@ -0,0 +1,86 @@
...
+  static Stream<Arguments> allPrimeInputs() {
+    int[] allPrimesUnder50 = { 2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47 };
+    Stream<Arguments> allPrimeInputs = Stream.of();
+    for (int primeNumber : allPrimesUnder50) {
+      allPrimeInputs = Stream.concat(allPrimeInputs, Stream.of(Arguments.of(primeNumber)));
+    }
+    return allPrimeInputs;
+  }
+
+  static Stream<Arguments> repetitiveInputs() {
```
Hedgemon4, line 45: Redo this with some hardcoded cases with negatives, integers greater than 50, and proper inputs after a varied amount of invalid ones.

src/loops/programs_with_repetition/easy/q4/MainTest.java
```diff
@@ -0,0 +1,86 @@
...
+        new StringLiteral("Entered a prime number that is less than 50. Ending Program. "),
+    });
+  }
+
+  @ParameterizedTest
+  @MethodSource("repetitiveInputs")
+  public void printsCorrectMessageWithMultipleInputs(ArrayList<Integer> inputs) throws InvalidClauseException {
+    TestOption.incorrectStructureErrorMessage = "Your program does not stop when it enters a prime number that is less than 50.";
+    Iterator<Integer> it = inputs.iterator();
+    Clause[][] outputStrings = new Clause[1][inputs.size() * 3 - 2];
```
Hedgemon4, line 77: For cleaner code, if I need to build a clause like this, I typically will put that in a separate private method for readability.

Hedgemon4 (changes requested):

src/loops/programs_with_repetition/easy/q4/ItsAlwaysPrimeTime.java
```diff
@@ -0,0 +1,30 @@
+package loops.programs_with_repetition.easy.q4;
+
+/*
+ * Write a program to keep asking the user to enter a number until they enter a prime number that is lesser than 50. Make sure your output matches the sample. 
+ * 
+ * Sample output:
+ *    Enter an integer:
+ *    2
```
Hedgemon4, line 8: Again, we should probably have some invalid input in the sample so the student can see that "Enter an integer:"
 is displayed multiple times.

src/loops/programs_with_repetition/easy/q4/MainTest.java
```diff
@@ -0,0 +1,107 @@
...
+    int[] allPrimesUnder50 = { 2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47 };
+    Stream<Arguments> allPrimeInputs = Stream.of();
+    for (int primeNumber : allPrimesUnder50) {
+      allPrimeInputs = Stream.concat(allPrimeInputs, Stream.of(Arguments.of(primeNumber)));
+    }
+    return allPrimeInputs;
+  }
+
+  private static Stream<Arguments> repetitiveHardcodedInputs() {
+    return Stream.of(
```
Hedgemon4, line 46: There should also be a prime number in one of these samples that is greater than 50 so we know that does not slip through the program

Hedgemon4: If they forget to check that boundary.

Hedgemon4 (changes requested):

Hedgemon4 (changes requested):

src/loops/programs_with_repetition/easy/q4/ItsAlwaysPrimeTime.java
```diff
@@ -0,0 +1,36 @@
+package loops.programs_with_repetition.easy.q4;
+
+/*
+ * Write a program to keep asking the user to enter a number until they enter a prime number that is lesser than 50. Make sure your output matches the sample.
```
Hedgemon4, line 4:
```suggestion
 * Write a program to keep asking the user to enter a number until they enter a prime number that is less than 50. Make sure your output matches the sample. 
```

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #297, opened 2022-07-16
loops-simple_programs_with_repetition-hard-q5

SamamthaH:
## PR Checklist
- [ ]  EOF newlines
---
## Website

### DIFF
--- a/src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
+++ b/src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
@@ -7,6 +7,7 @@ import global.variables.Clause;
 import global.variables.clauses.NewLine;
 import global.variables.clauses.PlaceHolder;
 import global.variables.clauses.StringLiteral;
+import global.variables.wrappers.Optional;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
@@ -19,7 +20,8 @@ public class MainTest extends BaseTest {
         TestOption.isInputTest = true;
         TestOption.defaultInput = "6";
         return new Clause[]{
-                new StringLiteral("Enter a number: "),
+                new StringLiteral("Enter a number:"),
+                new Optional(new StringLiteral(" ")),
                 new NewLine(),
                 new PlaceHolder()
         };
--- a/src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
+++ b/src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
@@ -35,7 +35,7 @@ public class MainTest extends BaseTest {
 
     @ParameterizedTest
     @MethodSource("inputProvider")
-    void discoversPerfectNumberCorrectly(String in, String place) throws InvalidClauseException {
+    void identifiesPerfectNumbersCorrectly(String in, String place) throws InvalidClauseException {
         TestOption.incorrectStructureErrorMessage = "Your program does not correctly identify perfect numbers.";
         runWithInput(in, new Clause[]{
                 new StringLiteral(place)
--- a/src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
+++ b/src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
@@ -1,11 +1,17 @@
 package loops.simple_programs_with_repitition.hard.q5;
 
 import global.BaseTest;
+import global.exceptions.InvalidClauseException;
 import global.tools.TestOption;
 import global.variables.Clause;
 import global.variables.clauses.NewLine;
 import global.variables.clauses.PlaceHolder;
 import global.variables.clauses.StringLiteral;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
 
 public class MainTest extends BaseTest {
     // Java
@@ -20,4 +26,17 @@ public class MainTest extends BaseTest {
     }
 
     public void runMain(){ NowThatsPerfect.main(new String[0]); }
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of("6", "6 is a Perfect Number!"), Arguments.of("7", "7 is NOT a Perfect Number!"), Arguments.of("-3", "-3 is NOT a Perfect Number!"), Arguments.of("0", "0 is a Perfect Number!"), Arguments.of("33550336", "33550336 is a Perfect Number!"));
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void discoversPerfectNumberCorrectly(String in, String place) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Either you have not correctly identified Perfect Numbers or your formatting is incorrect. Please refer to the sample provided.";
+        runWithInput(in, new Clause[]{
+                new StringLiteral(place)
+        });
+    }
 }
--- a/src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
+++ b/src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
@@ -30,7 +30,7 @@ public class MainTest extends BaseTest {
     public void runMain(){ NowThatsPerfect.main(new String[0]); }
 
     static Stream<Arguments> inputProvider(){
-        return Stream.of(Arguments.of("6", "6 is a Perfect Number!"), Arguments.of("7", "7 is NOT a Perfect Number!"), Arguments.of("-3", "-3 is NOT a Perfect Number!"), Arguments.of("0", "0 is NOT a Perfect Number!"), Arguments.of("33550336", "33550336 is a Perfect Number!", Arguments.of("546", "546 is NOT a Perfect Number!"), Arguments.of("8128", "8128 is a Perfect Number!"), Arguments.of("-6", "-6 is NOT a Perfect Number!")));
+        return Stream.of(Arguments.of("-6", "-6 is NOT a Perfect Number!"), Arguments.of("546", "546 is NOT a Perfect Number!"), Arguments.of("8128", "8128 is a Perfect Number!"), Arguments.of("1", "1 is NOT a Perfect Number!"), Arguments.of("6", "6 is a Perfect Number!"), Arguments.of("7", "7 is NOT a Perfect Number!"), Arguments.of("-3", "-3 is NOT a Perfect Number!"), Arguments.of("0", "0 is NOT a Perfect Number!"), Arguments.of("33550336", "33550336 is a Perfect Number!"));
     }
 
     @ParameterizedTest
--- a/src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
+++ b/src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
@@ -34,7 +34,7 @@ public class MainTest extends BaseTest {
     @ParameterizedTest
     @MethodSource("inputProvider")
     void discoversPerfectNumberCorrectly(String in, String place) throws InvalidClauseException {
-        TestOption.incorrectStructureErrorMessage = "Either you have not correctly identified Perfect Numbers or your formatting is incorrect. Please refer to the sample provided.";
+        TestOption.incorrectStructureErrorMessage = "Your program has not correctly identified Perfect Numbers.";
         runWithInput(in, new Clause[]{
                 new StringLiteral(place)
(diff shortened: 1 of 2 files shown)


SamamthaH:
I am unsure if the tests I wrote can really stop someone from hardcoding this question. There are only 5 perfect numbers that fit within Java's integer range so if someone does a quick google search all they need is to grab the input and check it against a short if statement. Not sure what I can do about it though.

Hedgemon4:
> I am unsure if the tests I wrote can really stop someone from hardcoding this question. There are only 5 perfect numbers that fit within Java's integer range so if someone does a quick google search all they need is to grab the input and check it against a short if statement. Not sure what I can do about it though.

Yeah, I don't think we can do much to prevent that, let me think about what we should do.

src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
```diff
@@ -0,0 +1,42 @@
...
+    public void runMain(){ NowThatsPerfect.main(new String[0]); }
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of("6", "6 is a Perfect Number!"), Arguments.of("7", "7 is NOT a Perfect Number!"), Arguments.of("-3", "-3 is NOT a Perfect Number!"), Arguments.of("0", "0 is a Perfect Number!"), Arguments.of("33550336", "33550336 is a Perfect Number!"));
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void discoversPerfectNumberCorrectly(String in, String place) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Either you have not correctly identified Perfect Numbers or your formatting is incorrect. Please refer to the sample provided.";
```
Hedgemon4, line 37: I think this only triggers if the PlaceHolder is wrong, and I think we should just say that their program does not correctly identify perfect numbers.

src/loops/simple_programs_with_repitition/hard/q5/NowThatsPerfect.java
```diff
@@ -0,0 +1,38 @@
+package loops.simple_programs_with_repitition.hard.q5;
+
+/*
+Write a program with the class name "NowThatsPerfect" to accept a number from the user and determine whether it is a "Perfect" number.
+(Hint: A perfect number is one whose factors sum up to equal the given number.
```
Hedgemon4, line 5: I think we should just define what a perfect number is instead of putting it in a hint (because honestly without the hint I would not know what a perfect number is)

src/loops/simple_programs_with_repitition/hard/q5/NowThatsPerfect.java
```diff
@@ -0,0 +1,38 @@
...
+Eg. 6 has factors of 1, 2, and 3, whose sum is 6, so it is a perfect number)
+
+Sample Output:
+Enter a number:
+5
+5 is NOT a Perfect Number!
+
+Note: Add an extra space after the colon in the output prompting the user.
```
Hedgemon4, line 13: To get around this, you can actually use `new Optional(new StringLiteral(" ")`

Hedgemon4: I do this so we don't accidentally penalize them for this

Hedgemon4 (changes requested):

src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
```diff
@@ -0,0 +1,44 @@
...
+                new Optional(new StringLiteral(" ")),
+                new NewLine(),
+                new PlaceHolder()
+        };
+    }
+
+    public void runMain(){ NowThatsPerfect.main(new String[0]); }
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of("6", "6 is a Perfect Number!"), Arguments.of("7", "7 is NOT a Perfect Number!"), Arguments.of("-3", "-3 is NOT a Perfect Number!"), Arguments.of("0", "0 is a Perfect Number!"), Arguments.of("33550336", "33550336 is a Perfect Number!"));
```
Hedgemon4, line 33: I would add one more multi-digit non-perfect number.

src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void discoversPerfectNumberCorrectly(String in, String place) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program has not correctly identified Perfect Numbers.";
```
Hedgemon4, line 39:
```suggestion
        TestOption.incorrectStructureErrorMessage = "Your program does not correctly indentify perfect numbers.";
```

Hedgemon4 (changes requested):

src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
Hedgemon4: And maybe one more perfect number. I think for java questions we should have more cases (up to a certain point)

Hedgemon4 (changes requested):

src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
Hedgemon4: You should change this error message to my recommended one.

src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
Hedgemon4: Add another larger non-perfect number

Hedgemon4 (changes requested):

src/loops/simple_programs_with_repitition/hard/q5/NowThatsPerfect.java
```diff
@@ -0,0 +1,36 @@
+package loops.simple_programs_with_repitition.hard.q5;
+
+/*
+Write a program with the class name "NowThatsPerfect" to accept a number from the user and determine whether it is a "Perfect" number.
+A perfect number is one whose factors sum up to equal the given number.
```
Hedgemon4, line 5: As this is a java question, we also need to also provide what they need to print if the number is a perfect number. I would just edit the question to say something like "If the number is a perfect number, print x is a Perfect Number!".

Hedgemon4 (changes requested):

src/loops/simple_programs_with_repitition/hard/q5/NowThatsPerfect.java
```diff
@@ -0,0 +1,36 @@
+package loops.simple_programs_with_repitition.hard.q5;
+
+/*
+Write a program with the class name "NowThatsPerfect" to accept a number from the user and determine whether it is a "Perfect" number.
+If it is a perfect number print: "x is a Perfect Number!". A perfect number is one whose factors sum up to equal the given number.
```
Hedgemon4, line 5: Even though it is in the sample output, I think what it should print if it is not a perfect number should also be specified in the question text.

Hedgemon4 (changes requested):

src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public void runMain(){ NowThatsPerfect.main(new String[0]); }
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of("6", "6 is a Perfect Number!"), Arguments.of("7", "7 is NOT a Perfect Number!"), Arguments.of("-3", "-3 is NOT a Perfect Number!"), Arguments.of("0", "0 is a Perfect Number!"), Arguments.of("33550336", "33550336 is a Perfect Number!", Arguments.of("546", "546 is NOT a Perfect Number!"), Arguments.of("8128", "8128 is a Perfect Number!")));
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void discoversPerfectNumberCorrectly(String in, String place) throws InvalidClauseException {
```
Hedgemon4, line 38:
```suggestion
    void identifiesPerfectNumbersCorrectly(String in, String place) throws InvalidClauseException {
```

Hedgemon4 (changes requested):

Hedgemon4 (approved):

src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
```diff
@@ -0,0 +1,44 @@
...
+                new Optional(new StringLiteral(" ")),
+                new NewLine(),
+                new PlaceHolder()
+        };
+    }
+
+    public void runMain(){ NowThatsPerfect.main(new String[0]); }
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of("6", "6 is a Perfect Number!"), Arguments.of("7", "7 is NOT a Perfect Number!"), Arguments.of("-3", "-3 is NOT a Perfect Number!"), Arguments.of("0", "0 is a Perfect Number!"), Arguments.of("33550336", "33550336 is a Perfect Number!", Arguments.of("546", "546 is NOT a Perfect Number!"), Arguments.of("8128", "8128 is a Perfect Number!")));
```
Hedgemon4, line 33: I think we need to change the question code because 0 should not be a perfect number.

Hedgemon4 (changes requested):

src/loops/simple_programs_with_repitition/hard/q5/NowThatsPerfect.java
```diff
@@ -0,0 +1,39 @@
...
+Eg. 6 has factors of 1, 2, and 3, whose sum is 6, so it is a perfect number.
+
+Sample Output:
+Enter a number:
+5
+5 is NOT a Perfect Number!
+ */
+
+import java.util.Scanner;
+public class NowThatsPerfect
```
Hedgemon4, line 16: This did not fix the problem because now the question will output that -1 is a perfect number.

Hedgemon4 (changes requested):

src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
```diff
@@ -0,0 +1,44 @@
...
+                new Optional(new StringLiteral(" ")),
+                new NewLine(),
+                new PlaceHolder()
+        };
+    }
+
+    public void runMain(){ NowThatsPerfect.main(new String[0]); }
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of("6", "6 is a Perfect Number!"), Arguments.of("7", "7 is NOT a Perfect Number!"), Arguments.of("-3", "-3 is NOT a Perfect Number!"), Arguments.of("0", "0 is NOT a Perfect Number!"), Arguments.of("33550336", "33550336 is a Perfect Number!", Arguments.of("546", "546 is NOT a Perfect Number!"), Arguments.of("8128", "8128 is a Perfect Number!")));
```
Hedgemon4, line 33: We should also add a case to this of a number that is a perfect number, but negative so it should return that it is not a Perfect Number.

Hedgemon4 (changes requested):

src/loops/simple_programs_with_repitition/hard/q5/MainTest.java
```diff
@@ -0,0 +1,44 @@
...
+                new Optional(new StringLiteral(" ")),
+                new NewLine(),
+                new PlaceHolder()
+        };
+    }
+
+    public void runMain(){ NowThatsPerfect.main(new String[0]); }
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of("6", "6 is a Perfect Number!"), Arguments.of("7", "7 is NOT a Perfect Number!"), Arguments.of("-3", "-3 is NOT a Perfect Number!"), Arguments.of("0", "0 is NOT a Perfect Number!"), Arguments.of("33550336", "33550336 is a Perfect Number!", Arguments.of("546", "546 is NOT a Perfect Number!"), Arguments.of("8128", "8128 is a Perfect Number!"), Arguments.of("-6", "-6 is NOT a Perfect Number!")));
```
Hedgemon4, line 33: I think we should add the case of the number 1 because I could see the student programming the loop incorrectly and allowing that to pass when it shouldn't.

Hedgemon4: Otherwise, I think this will be good to merge

Hedgemon4 (changes requested):

Hedgemon4 (changes requested):

Hedgemon4 (approved):

Issue #131, opened 2022-05-24
Add support for exponents to DoubleLiteral and RandomDouble


Hedgemon4:

Hedgemon4:
See https://github.com/canvas-gamification/canvas-gamification-junit-tests/pull/62

Opeyem1a:
This is going to be a regex thing with double literal and random double, maybe look online for a catch all double catcher (which solves the problem for random double), but supporting double literal means we need bounds

Hedgemon4:
This has been separated into two issues (see https://github.com/canvas-gamification/canvas-gamification-junit-tests/issues/317 and https://github.com/canvas-gamification/canvas-gamification-junit-tests/issues/316).

PR #311, opened 2022-07-16
fix double regex

Opeyem1a:
Closes half of #131


Hedgemon4 (approved):

Hedgemon4:
Double regex now matches exponents

PR #313, opened 2022-07-17
loops-programs_with_repetition-hard-q4

SamamthaH:
## PR Checklist
- [ ]  EOF newlines
---
## Website

### DIFF
--- a/src/loops/programs_with_repetition/hard/q4/ArtificialSelection.java
+++ b/src/loops/programs_with_repetition/hard/q4/ArtificialSelection.java
@@ -1,0 +1,34 @@
+package loops.programs_with_repetition.hard.q4;
+
+/*
+Write a program that accepts a sentence from the user and prints the sentence after reversing every other word,
+starting from the first word, using nested loops. Assume that every word is separated by a single space " ",
+and there are no consecutive spaces.
+ */
+
+import java.util.Scanner;
+
+public class ArtificialSelection {
+    public static void main(String[] args) {
+        Scanner input = new Scanner(System.in);
+        System.out.println("Enter a sentence: ");
+        String sen = " " + input.nextLine();
+        int spaceCount = 0;
+        for (int i = 0; i < sen.length(); i++) {
+            int nextSp = sen.indexOf(' ', i + 1);
+            if (nextSp == -1)
+                nextSp = sen.length();
+            if (sen.charAt(i) == ' ' && i != sen.length() - 1) {
+                spaceCount++;
+                String word = sen.substring(i + 1, nextSp);
+                if (spaceCount % 2 == 1)
+                    for (int j = word.length() - 1; j >= 0; j--)
+                        System.out.print(word.charAt(j));
+                else
+                    for (int j = 0; j < word.length(); j++)
+                        System.out.print(word.charAt(j));
+                System.out.print(" ");
+            }
+        }
+    }
+}
--- a/src/loops/programs_with_repetition/hard/q4/ArtificialSelection.java
+++ b/src/loops/programs_with_repetition/hard/q4/ArtificialSelection.java
@@ -2,7 +2,7 @@ package loops.programs_with_repetition.hard.q4;
 
 /*
 Write a program that accepts a sentence from the user and prints the sentence after reversing every other word,
-starting from the first word, using nested loops. Assume that every word is separated by a single space " ",
+starting with the first word, using nested loops. Assume that every word is separated by a single space " ",
 and there are no consecutive spaces.
 
 Sample Output:
--- a/src/loops/programs_with_repetition/hard/q4/ArtificialSelection.java
+++ b/src/loops/programs_with_repetition/hard/q4/ArtificialSelection.java
@@ -4,6 +4,11 @@ package loops.programs_with_repetition.hard.q4;
 Write a program that accepts a sentence from the user and prints the sentence after reversing every other word,
 starting from the first word, using nested loops. Assume that every word is separated by a single space " ",
 and there are no consecutive spaces.
+
+Sample Output:
+Enter a sentence:
+Which wrist watch is a Swiss wrist watch
+hcihW wrist hctaw is a Swiss tsirw watch
  */
 
 import java.util.Scanner;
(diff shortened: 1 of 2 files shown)


src/loops/programs_with_repetition/hard/q4/ArtificialSelection.java
```diff
@@ -0,0 +1,39 @@
+package loops.programs_with_repetition.hard.q4;
+
+/*
+Write a program that accepts a sentence from the user and prints the sentence after reversing every other word,
+starting from the first word, using nested loops. Assume that every word is separated by a single space " ",
```
Hedgemon4, line 5:
```suggestion
starting with the first word, using nested loops. Assume that every word is separated by a single space " ",
```

src/loops/programs_with_repetition/hard/q4/MainTest.java
```diff
@@ -0,0 +1,46 @@
...
+                new PlaceHolder(),
+                new Optional(new StringLiteral(" "))
+        };
+    }
+
+    public void runMain(){ArtificialSelection.main(new String[0]);}
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of("warning, they are coming to find you", ",gninraw they era coming ot find uoy"),
+                Arguments.of("hello", "olleh"), Arguments.of("if i put a bunch of palindromes in here it would print out the same thing, wouldn't that be so pointless", "fi i tup a hcnub of semordnilap in ereh it dluow print tuo the emas thing, t'ndluow that eb so sseltniop "));
```
Hedgemon4, line 35: What happens if you use special characters, capitalization, numbers, etc?

src/loops/programs_with_repetition/hard/q4/MainTest.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void reversesEveryOtherCorrectly(String in, String reversed) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not correctly reverse every other word according to the instruction.";
```
Hedgemon4, line 41:
```suggestion
        TestOption.incorrectStructureErrorMessage = "Your program does not correctly reverse every other word.";
```

src/loops/programs_with_repetition/hard/q4/MainTest.java
Hedgemon4: I would add another test case involving those (if it doesn't break anything).

Hedgemon4 (changes requested):

src/loops/programs_with_repetition/hard/q4/MainTest.java
```diff
@@ -0,0 +1,56 @@
...
+    @MethodSource("inputProvider")
+    void reversesEveryOtherCorrectly(String in, String reversed) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not correctly reverse every other word.";
+        runWithInput(in, new Clause[]{
+                new StringLiteral(reversed)
+        });
+    }
+
+    @Test
+    void handlesSpecialCharacters() throws InvalidClauseException {
```
Hedgemon4, line 50: I think this is confusing because we don't explicitly say to handle special characters differently in the question. I just mean you should try to include a few in your test ( like ending a sentence with an exclamation mark ).

Hedgemon4: So I would just put some special characters in the test cases you already have and then you should be good to go

src/loops/programs_with_repetition/hard/q4/MainTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of("warning, they are coming to find you", ",gninraw they era coming ot find uoy"),
+                Arguments.of("hello", "olleh"), Arguments.of("if i put a bunch of palindromes in here it would print out the same thing, wouldn't that be so pointless", "fi i tup a hcnub of semordnilap in ereh it dluow print tuo the emas thing, t'ndluow that eb so sseltniop "),
+                Arguments.of("There's A Storm Still COMING", "s'erehT A mrotS Still GNIMOC "), Arguments.of("23,000 leagues under the sea", "000,32 leagues rednu the aes "), Arguments.of("W!~!! !~!~!~~ ~~~~!!!", "!!~!W !~!~!~~ !!!~~~~"));
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void reversesEveryOtherCorrectly(String in, String reversed) throws InvalidClauseException {
```
Hedgemon4, line 42:
```suggestion
    void reversesEveryOtherWord(String in, String reversed) throws InvalidClauseException {
```

Hedgemon4 (changes requested):

src/loops/programs_with_repetition/hard/q4/ArtificialSelection.java
```diff
@@ -0,0 +1,39 @@
+package loops.programs_with_repetition.hard.q4;
+
+/*
+Write a program that accepts a sentence from the user and prints the sentence after reversing every other word,
+starting with the first word, using nested loops. Assume that every word is separated by a single space " ",
+and there are no consecutive spaces.
+
```
Hedgemon4, line 7: What would you think about adding a hint that there should be a space before the first word? I think it is not necessarily obvious here.

Hedgemon4: I just worry that people would get quite stuck without it, and they might assume our test contains a space at the start.

Hedgemon4: Or go about it logically in another way which would make it difficult to pass our test, even if they went about it correctly.

SamamthaH: I don't know, the way I see it, even if the students go about it a different way the test should still work, that's why we treat java questions like a black box we can't see into: put something in and see what comes out without seeing the process. An easy way to get around the first space is for them to just flip the first word outside the loops and then start the loops after in my opinion, but admittedly I haven't tried to solve the question myself. I don't see why they would assume our test contains a space at the start.

Hedgemon4: I was just worried that someone could assume that if every word is separated by space, they might think that includes having one in front of the first word. However, that is probably an edge case, so

Hedgemon4 (approved):

PR #314, opened 2022-07-17
loops-programs_with_repetition-hard-q6

SamamthaH:
## PR Checklist
- [ ]  EOF newlines
---
## Website

### DIFF
--- a/src/loops/programs_with_repetition/hard/q6/KingOfTheString.java
+++ b/src/loops/programs_with_repetition/hard/q6/KingOfTheString.java
@@ -3,6 +3,11 @@ package loops.programs_with_repetition.hard.q6;
 /*
 Write a program named KingOfTheString that will ask the user for a sentence and then display the most common character in the given sentence.
 You can ignore the case of the characters in the sentence.
+
+Sample Output:
+Enter a sentence:
+Hello Tallulah like the lovely weather?
+l 8
  */
 
 import java.util.Scanner;
@@ -13,7 +18,7 @@ public class KingOfTheString
         Scanner input = new Scanner( System.in );
         System.out.println( "Enter a sentence: " );
         String sent = input.nextLine();
-        sent.toLowerCase();
+        sent = sent.toLowerCase();
         char maxc = ' ';
         int maxn = 0;
         for(int i = 0; i < sent.length(); i++)
--- a/src/loops/programs_with_repetition/hard/q6/KingOfTheString.java
+++ b/src/loops/programs_with_repetition/hard/q6/KingOfTheString.java
@@ -1,0 +1,33 @@
+package loops.programs_with_repetition.hard.q6;
+
+/*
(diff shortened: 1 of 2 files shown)


src/loops/programs_with_repetition/hard/q6/MainTest.java
```diff
@@ -0,0 +1,44 @@
...
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of("Hello Tallulah like the lovely weather", "l 8"), Arguments.of("IIii aaa", "i 4"), Arguments.of(" ", "  1"), Arguments.of("!!!! dumdundundun !!!!!!", "! 10"));
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void countsMostUsedCharacterCorrectly(String in, String count) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not correctly identify and count the most common character.";
+        runWithInput(in, new Clause[]{
+                new StringLiteral(count)
```
Hedgemon4, line 41: I would use a clause and IntegerLiteral here

Hedgemon4: That way, you could give separate feedback for identifying the correct letter, and for counting incorrectly.

Hedgemon4 (changes requested):

src/loops/programs_with_repetition/hard/q6/KingOfTheString.java
```diff
@@ -0,0 +1,38 @@
+package loops.programs_with_repetition.hard.q6;
+
+/*
+Write a program named KingOfTheString that will ask the user for a sentence and then display the most common character in the given sentence.
+You can ignore the case of the characters in the sentence. When you print the most common character print it in lower case.
```
Hedgemon4, line 5:
```suggestion
Ignore the case of the characters in the sentence. When you print the most common character print it in lower case.
```

Hedgemon4 (approved):

src/loops/programs_with_repetition/hard/q6/MainTest.java
```diff
@@ -0,0 +1,51 @@
...
+                new Optional(new StringLiteral(" ")),
+                new NewLine(),
+                new PlaceHolder()
+        };
+    }
+
+    public void runMain(){KingOfTheString.main(new String[0]);}
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of("Hello Tallulah like the lovely weather", "l", "8"), Arguments.of("IIii aaa", "i", "4"), Arguments.of(" ", " ", "1"), Arguments.of("!!!! dumdundundun !!!!!!", "!", "10"));
```
Hedgemon4, line 37: I think you should add one more general case that is not the sample string.

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #315, opened 2022-07-17
loops-programs_with_repetition-hard-q7

SamamthaH:
## PR Checklist
- [ ]  EOF newlines
---
## Website

### DIFF
--- a/src/loops/programs_with_repetition/hard/q7/MainTest.java
+++ b/src/loops/programs_with_repetition/hard/q7/MainTest.java
@@ -12,7 +12,9 @@ public class MainTest extends BaseTest {
     public Clause[] testSentence(){
         TestOption.isInputTest = false;
         return new Clause[]{
-                new StringLiteral("Approximation of sin(3.14) using n = "),
+                new StringLiteral("Approximation of sin\\("),
+                new DoubleLiteral("x"),
+                new StringLiteral("\\) using n = "),
                 new IntegerLiteral("num"),
                 new StringLiteral(" is "),
                 new DoubleLiteral("approx")
--- a/src/loops/programs_with_repetition/hard/q7/MainTest.java
+++ b/src/loops/programs_with_repetition/hard/q7/MainTest.java
@@ -5,20 +5,31 @@ import global.tools.TestOption;
 import global.variables.Clause;
 import global.variables.clauses.DoubleLiteral;
 import global.variables.clauses.IntegerLiteral;
+import global.variables.clauses.NewLine;
 import global.variables.clauses.StringLiteral;
+import global.variables.wrappers.Optional;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 public class MainTest extends BaseTest {
     // Java
     public Clause[] testSentence(){
-        TestOption.isInputTest = false;
+        TestOption.isInputTest = true;
+        TestOption.defaultInput = "3.14";
         return new Clause[]{
+                new StringLiteral("Enter a number to approximate:"),
+                new Optional(new StringLiteral(" ")),
+                new NewLine(),
                 new StringLiteral("Approximation of sin\\("),
                 new DoubleLiteral("x"),
                 new StringLiteral("\\) using n = "),
-                new IntegerLiteral("n"),
+                new IntegerLiteral(5),
                 new StringLiteral(" is "),
                 new DoubleLiteral("approx")
         };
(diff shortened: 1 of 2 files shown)


src/loops/programs_with_repetition/hard/q7/SinOfTheTimes.java
```diff
@@ -0,0 +1,31 @@
+package loops.programs_with_repetition.hard.q7;
+
+/*
+Write a program called SinOfTheTimes that will calculate an approximation of sin(x) from n = 0 until n = 10, with x = 3.14,
```
Hedgemon4, line 4: I think we should change this question to specify one of the two numbers, so they can't hardcode it.

Hedgemon4: And then we can vary the other one on the website

Hedgemon4 (changes requested):

src/loops/programs_with_repetition/hard/q7/SinOfTheTimes.java
Hedgemon4: Yeah, but if we change the values, they could still just use something like wolfram alpha to quickly calculate the series, and then just print that out.

Hedgemon4 (changes requested):

SamamthaH:
I've changed it over to an input test but in doing so realized there's actually a problem with running this program with java's limited int size. We can't make n greater than 5 because afterwards the factorials get too big for java integers to handle, and we can't approximate higher than 4.4 without the approximation going higher than 1 or lower than -1 for an unknown reason. Because the n is so limited people won't need to use loops, they can just hardcode in 5 lines of the equation so I don't know if this can work as an actual question.

src/loops/programs_with_repetition/hard/q7/SinOfTheTimes.java
```diff
@@ -0,0 +1,34 @@
+package loops.programs_with_repetition.hard.q7;
+
+/*
+Write a program called SinOfTheTimes that will calculate an approximation of sin(x) from n = 0 until n = 5, with x = 3.14,
```
Hedgemon4, line 4: This text is not up to date with the question being changed to have an input.

src/loops/programs_with_repetition/hard/q7/SinOfTheTimes.java
```diff
@@ -0,0 +1,34 @@
...
+using the following series ( n! symbolises factorials, where n! = n (n - 1)( n - 2 )...(1) ):
+x - ( x^3 / 3! ) + ( x^5 / 5! ) - ( x^7 / 7! ) + … + ((-1)^n x^(2n + 1))/( 2n + 1)!
+
+Sample Output:
+	Approximation of sin(3.14) using n = 5 is …
```
Hedgemon4, line 9: The sample output is also out of date.

Hedgemon4 (changes requested):

Hedgemon4:
> I've changed it over to an input test but in doing so realized there's actually a problem with running this program with java's limited int size. We can't make n greater than 5 because afterwards the factorials get too big for java integers to handle, and we can't approximate higher than 4.4 without the approximation going higher than 1 or lower than -1 for an unknown reason. Because the n is so limited people won't need to use loops, they can just hardcode in 5 lines of the equation so I don't know if this can work as an actual question.

This can be solved by changing the calculation to use double values instead of integer ones.

src/loops/programs_with_repetition/hard/q7/SinOfTheTimes.java
```diff
@@ -0,0 +1,36 @@
...
+
+/*
+Write a program called SinOfTheTimes that will take an input x and use it to calculate an approximation of sin(x)
+from n = 0 until n = 10 using the following series ( n! symbolises factorials, where n! = n (n - 1)( n - 2 )...(1) ):
+x - ( x^3 / 3! ) + ( x^5 / 5! ) - ( x^7 / 7! ) + … + ((-1)^n x^(2n + 1))/( 2n + 1)! Calculate using doubles due to the size limit of integers.
+
+Sample Output:
+    Enter a number to approximate:
+    3.14
+	Approximation of sin(3.14) using n = 10 is …
```
Hedgemon4, line 11:
```suggestion
    Approximation of sin(3.14) using n = 10 is x
```

src/loops/programs_with_repetition/hard/q7/MainTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of(2.56, "0.5493554364272203"), Arguments.of(1.07, "0.8772005042746817"), Arguments.of(4.2, "-0.8715757642927082"), Arguments.of(24, "1.052000954450763E9"));
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void correctApproximationCalculated(double in, String approx){
+        runWithInput(String.valueOf(in));
+        assertEquals(approx, getItemByName("approx"), "Your approximation is incorrect.");
```
Hedgemon4, line 47:
```suggestion
        assertEquals(approx, getItemByName("approx"), "Your approximation of sin(x) is incorrect.");
```

Hedgemon4, line 47:
```suggestion
        assertEquals(approx, getItemByName("approx"), "Your approximation of sin(x) is incorrect.");
```

src/loops/programs_with_repetition/hard/q7/MainTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public void runMain(){SinOfTheTimes.main(new String[0]);}
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of(2.56, "0.5493554364272203"), Arguments.of(1.07, "0.8772005042746817"), Arguments.of(4.2, "-0.8715757642927082"), Arguments.of(24, "1.052000954450763E9"));
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void correctApproximationCalculated(double in, String approx){
```
Hedgemon4, line 45:
```suggestion
    void calculatesApproximationCorrectly(double in, String approx){
```

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #320, opened 2022-07-17
loops-nested_loops-hard-q6

SamamthaH:
…ode.

## PR Checklist
- [ ]  EOF newlines
---
## Website

### DIFF
--- a/src/loops/nested_loops/hard/q6/MainTest.java
+++ b/src/loops/nested_loops/hard/q6/MainTest.java
@@ -1,0 +1,57 @@
+package loops.nested_loops.hard.q6;
+
+import global.BaseTest;
+import global.exceptions.InvalidClauseException;
+import global.tools.TestOption;
+import global.variables.Clause;
+import global.variables.clauses.NewLine;
+import global.variables.clauses.PlaceHolder;
+import global.variables.clauses.StringLiteral;
+import global.variables.wrappers.Optional;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
+
+public class MainTest extends BaseTest {
+    // Java
+    public Clause[] testSentence(){
+        TestOption.isInputTest = true;
+        TestOption.defaultInput = "5";
+        return new Clause[]{
+                new StringLiteral("Enter number of lines:"),
+                new Optional(new StringLiteral(" ")),
+                new NewLine(),
+                new PlaceHolder()
+        };
+    }
+
+    public void runMain(){TryAngleButBothWaysAKAPyramid.main(new String[0]);}
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of(5), Arguments.of(1), Arguments.of(0), Arguments.of(10));
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void printsPyramidCorrectly(int n) throws InvalidClauseException {
+        Clause[][] place = new Clause[1][n*3];
+        int count = 0;
+        for(int x = 0; x <n; x++){
+            StringBuilder s = new StringBuilder();
+            s.append(" ".repeat(n - x - 1));
+            place[0][count++] = new StringLiteral(s.toString());
+            s = new StringBuilder();
+            for(int y = x+1; y > 0; y--){
+                s.append(y);
+            }
+            for(int y = 2; y <= x+1; y++){
+                s.append(y);
+            }
+            place[0][count++] = new StringLiteral(s.toString());
+            place[0][count++] = new NewLine();
+        }
+        runWithInput(String.valueOf(n), place);
+    }
+}
--- a/src/loops/nested_loops/hard/q6/MainTest.java
+++ b/src/loops/nested_loops/hard/q6/MainTest.java
@@ -36,6 +36,11 @@ public class MainTest extends BaseTest {
     @ParameterizedTest
     @MethodSource("inputProvider")
     void printsPyramidCorrectly(int n) throws InvalidClauseException {
+
+        runWithInput(String.valueOf(n), clauseBuilder(n));
+    }
+
+    Clause[][] clauseBuilder(int n){
         Clause[][] place = new Clause[1][n*3];
         int count = 0;
         for(int x = 0; x <n; x++){
@@ -52,6 +57,6 @@ public class MainTest extends BaseTest {
             place[0][count++] = new StringLiteral(s.toString());
             place[0][count++] = new NewLine();
         }
-        runWithInput(String.valueOf(n), place);
+        return place;
     }
 }
--- a/src/loops/nested_loops/hard/q6/MainTest.java
+++ b/src/loops/nested_loops/hard/q6/MainTest.java
@@ -36,7 +36,7 @@ public class MainTest extends BaseTest {
     @ParameterizedTest
     @MethodSource("inputProvider")
     void printsPyramidCorrectly(int n) throws InvalidClauseException {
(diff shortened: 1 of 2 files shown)


src/loops/nested_loops/hard/q6/MainTest.java
```diff
@@ -0,0 +1,57 @@
...
+    public void runMain(){TryAngleButBothWaysAKAPyramid.main(new String[0]);}
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(Arguments.of(5), Arguments.of(1), Arguments.of(0), Arguments.of(10));
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void printsPyramidCorrectly(int n) throws InvalidClauseException {
+        Clause[][] place = new Clause[1][n*3];
```
Hedgemon4, line 39: This is fine, but I think it is better form to put this in a method called clauseBuilder()

Hedgemon4: I just think it makes for cleaner code this way

src/loops/nested_loops/hard/q6/MainTest.java
```diff
@@ -0,0 +1,57 @@
...
+                new StringLiteral("Enter number of lines:"),
+                new Optional(new StringLiteral(" ")),
+                new NewLine(),
+                new PlaceHolder()
+        };
+    }
+
+    public void runMain(){TryAngleButBothWaysAKAPyramid.main(new String[0]);}
+
+    static Stream<Arguments> inputProvider(){
```
Hedgemon4, line 32: Turn this into a `Stream<Integer>` as that is all you are returning

src/loops/nested_loops/hard/q6/TryAngleButBothWaysAKAPyramid.java
```diff
@@ -0,0 +1,38 @@
...
+import global.exceptions.InvalidClauseException;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.Scanner;
+public class TryAngleButBothWaysAKAPyramid
+{
+    public static void main( String[] args )
+    {
+        Scanner input = new Scanner( System.in );
```
Hedgemon4, line 24: I would also add a check to make sure the user inputs a positive number, and it should print `Invalid input!` otherwise.

src/loops/nested_loops/hard/q6/MainTest.java
```diff
@@ -0,0 +1,57 @@
...
+        return Stream.of(Arguments.of(5), Arguments.of(1), Arguments.of(0), Arguments.of(10));
```
Hedgemon4, line 33: Also, you could test a much larger value, like 76

Hedgemon4 (changes requested):

src/loops/nested_loops/hard/q6/MainTest.java
```diff
@@ -0,0 +1,69 @@
...
+                s.append(y);
+            }
+            place[0][count++] = new StringLiteral(s.toString());
+            place[0][count++] = new NewLine();
+        }
+        return place;
+    }
+
+    @Test
+    void identifiesInvalidInput() throws InvalidClauseException {
```
Hedgemon4, line 63: This should have more cases, because since this is a java question if they made an off by one error something like -1 would pass this code when it should not. So you should test -1, and then one or two other negative values.

src/loops/nested_loops/hard/q6/TryAngleButBothWaysAKAPyramid.java
```diff
@@ -0,0 +1,37 @@
...
+        System.out.println("Enter number of lines: ");
+        int n = input.nextInt();
+        if(n <0)
```
Hedgemon4, line 23:
```suggestion
        if(n < 0)
```

Hedgemon4 (changes requested):

src/loops/nested_loops/hard/q6/MainTest.java
```diff
@@ -0,0 +1,72 @@
...
+    public void runMain(){TryAngleButBothWaysAKAPyramid.main(new String[0]);}
+
+    static Stream<Integer> inputProvider(){
+        return Stream.of(5, 1, 0, 10, 42);
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void printsPyramidCorrectly(int n) throws InvalidClauseException {
+        runWithInput(String.valueOf(n), clauseBuilder(n));
```
Hedgemon4, line 39: This is missing an error message.

src/loops/nested_loops/hard/q6/TryAngleButBothWaysAKAPyramid.java
```diff
@@ -0,0 +1,37 @@
+package loops.nested_loops.hard.q6;
+
+/*
+Write a program to ask the use for n ("Enter number of lines: "), and print the following pattern (Pyramid) using nested loops based on their input:
```
Hedgemon4, line 4:
```suggestion
Write a program to ask the user for n ("Enter number of lines: "), and print the following pyramid pattern using nested loops based on their input:
```

src/loops/nested_loops/hard/q6/TryAngleButBothWaysAKAPyramid.java
```diff
@@ -0,0 +1,37 @@
...
+           1
+          212
+         32123
+        4321234
+       543212345
+         … 1 ...
+n …   ...  1 … …  n
+If a negative integer is input print "Invalid input!".
+ */
+
```
Hedgemon4, line 14: This is also missing a sample output

Hedgemon4 (changes requested):

src/loops/nested_loops/hard/q6/MainTest.java
SamamthaH: I also figured here that the entire program was just formatting, I'm not even sure what an error message would be. Your pyramid was not built properly?

Hedgemon4: I mean it is a formatting issue, but I feel like the generic message should mostly be for if they have in error in a print statement or something, whereas here we can at least specify that they printed the pattern incorrectly, and not something else

src/loops/nested_loops/hard/q6/TryAngleButBothWaysAKAPyramid.java
```diff
@@ -0,0 +1,45 @@
+package loops.nested_loops.hard.q6;
+
+/*
+Write a program to ask the user for n ("Enter number of lines: "), and print the following pyramid pattern using
```
Hedgemon4, line 4:
```suggestion
Write a program to ask the user for an n ("Enter number of lines: "), and print the following pyramid pattern using
```

src/loops/nested_loops/hard/q6/MainTest.java
```diff
@@ -0,0 +1,73 @@
...
+    public void runMain(){TryAngleButBothWaysAKAPyramid.main(new String[0]);}
+
+    static Stream<Integer> inputProvider(){
+        return Stream.of(5, 1, 0, 10, 42);
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void printsPyramidCorrectly(int n) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program printed the pattern incorrectly.";
```
Hedgemon4, line 39:
```suggestion
        TestOption.incorrectStructureErrorMessage = "Your does not correctly print the triangle pattern.";
```

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #321, opened 2022-07-18
loops-nested_loops-hard-q7

SamamthaH:
## PR Checklist
- [ ]  EOF newlines
---
## Website

### DIFF
--- a/src/loops/nested_loops/hard/q7/BoxMeNow.java
+++ b/src/loops/nested_loops/hard/q7/BoxMeNow.java
@@ -2,7 +2,7 @@ package loops.nested_loops.hard.q7;
 
 /*
 Write a program called BoxMeNow to ask the use for n, and print the following pattern using nested loops based on their
-input.
+input. If the input is negative print “Invalid input!”
 Sample Output for n =  4:
 Enter number of lines:
 4
@@ -21,16 +21,18 @@ public class BoxMeNow
         Scanner input = new Scanner( System.in );
         System.out.println("Enter number of lines: ");
         int n = input.nextInt();
-        for(int i = 1; i <= n; i++)
-        {
-            for(int j = 1; j <= n; j++)
-            {
-                if( ( i == 1 || i == n ) || ( j == 1 || j == n ) )
-                    System.out.print( "1" );
-                else
-                    System.out.print( " " );
+        if(n < 0)
+            System.out.println("Invalid input!");
+        else {
+            for (int i = 1; i <= n; i++) {
+                for (int j = 1; j <= n; j++) {
+                    if ((i == 1 || i == n) || (j == 1 || j == n))
+                        System.out.print("1");
+                    else
+                        System.out.print(" ");
+                }
+                System.out.println();
             }
-            System.out.println();
         }
     }
 }
--- a/src/loops/nested_loops/hard/q7/BoxMeNow.java
+++ b/src/loops/nested_loops/hard/q7/BoxMeNow.java
@@ -1,9 +1,9 @@
 package loops.nested_loops.hard.q7;
 
 /*
-Write a program called BoxMeNow to ask the use for n, and print the following pattern using nested loops based on their
-input. If the input is less than zero print “Invalid input!”.
-Sample Output for n =  4:
+Write a program called BoxMeNow to ask the user for an n, and print the following pattern using nested loops,
+with the n being the side length of the box (the number of 1’s per side). If the input is less than zero print “Invalid input!”.
+
 Enter number of lines:
 4
 1 1 1 1
--- a/src/loops/nested_loops/hard/q7/BoxMeNow.java
+++ b/src/loops/nested_loops/hard/q7/BoxMeNow.java
@@ -2,7 +2,7 @@ package loops.nested_loops.hard.q7;
 
 /*
 Write a program called BoxMeNow to ask the use for n, and print the following pattern using nested loops based on their
-input. If the input is negative print “Invalid input!”
+input. If the input is less than zero print “Invalid input!”
 Sample Output for n =  4:
 Enter number of lines:
 4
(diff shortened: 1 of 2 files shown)


src/loops/nested_loops/hard/q7/BoxMeNow.java
```diff
@@ -0,0 +1,39 @@
+package loops.nested_loops.hard.q7;
+
+/*
+Write a program called BoxMeNow to ask the use for n, and print the following pattern using nested loops based on their
+input. If the input is less than zero print “Invalid input!”
```
Hedgemon4, line 5:
```suggestion
input. If the input is less than zero print “Invalid input!”.
```

src/loops/nested_loops/hard/q7/MainTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void constructPatternCorrectly(int n) throws InvalidClauseException {
+        runWithInput(String.valueOf(n), clauseBuilder(n));
+    }
+
+    @Test
+    void identifiesInvalidInput() throws InvalidClauseException{
+        runWithInput("-49", new Clause[]{
```
Hedgemon4, line 44: See comments about boundary checking on the previous question.

src/loops/nested_loops/hard/q7/BoxMeNow.java
```diff
@@ -0,0 +1,39 @@
...
+Sample Output for n =  4:
+Enter number of lines:
```
Hedgemon4, line 7: This should probably be the side length of the box, as we don't want anyone to get confused and only print four at the top.

src/loops/nested_loops/hard/q7/BoxMeNow.java
```diff
@@ -0,0 +1,39 @@
...
+4
+1111
```
Hedgemon4, line 9: I think we should change the pattern to have a space between the ones so it actually will stay a box

Hedgemon4: ie `1 1 1 1`

src/loops/nested_loops/hard/q7/BoxMeNow.java
```diff
@@ -0,0 +1,39 @@
+package loops.nested_loops.hard.q7;
+
+/*
+Write a program called BoxMeNow to ask the use for n, and print the following pattern using nested loops based on their
```
Hedgemon4, line 4:
```suggestion
Write a program called BoxMeNow to ask the user for an n, and print the following box pattern using nested loops, with the n being the side length of the box (the number of 1's per side). 
```

Hedgemon4 (changes requested):

src/loops/nested_loops/hard/q7/BoxMeNow.java
SamamthaH: I'm confused. The number is the number on the sides of the box and the sample output shows it.

src/loops/nested_loops/hard/q7/MainTest.java
```diff
@@ -0,0 +1,87 @@
...
+    public void runMain(){BoxMeNow.main(new String[0]);}
+
+    static Stream<Integer> inputProvider(){
+        return Stream.of(1, 0, 5, 23, 2);
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void constructPatternCorrectly(int n) throws InvalidClauseException {
+        runWithInput(String.valueOf(n), clauseBuilder(n));
```
Hedgemon4, line 38: This is missing an error message

src/loops/nested_loops/hard/q7/MainTest.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    static Stream<Integer> invalidInputProvider(){return Stream.of(-1, -47, -300);}
+
+    @ParameterizedTest
+    @MethodSource("invalidInputProvider")
+    void identifiesInvalidInput(int in) throws InvalidClauseException{
+        runWithInput(String.valueOf(in), new Clause[]{
```
Hedgemon4, line 46: This is missing an error message

Hedgemon4 (changes requested):

src/loops/nested_loops/hard/q7/MainTest.java
SamamthaH: I figure the standard wrong format would work here since the program is all format.

Hedgemon4: Same as the previous pull request, I think it is basically always better to specify an error message when possible.

Hedgemon4: It just allows us to target the feedback towards the things we test for, as opposed to if they just spell a word or something wrong.

Hedgemon4 (changes requested):

src/loops/nested_loops/hard/q7/MainTest.java
```diff
@@ -0,0 +1,89 @@
...
+    public void runMain(){BoxMeNow.main(new String[0]);}
+
+    static Stream<Integer> inputProvider(){
+        return Stream.of(1, 0, 5, 23, 2);
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void constructPatternCorrectly(int n) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not print the box correctly.";
```
Hedgemon4, line 38:
```suggestion
        TestOption.incorrectStructureErrorMessage = "Your program does not print the box pattern correctly.";
```

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #325, opened 2022-07-18
loops-nested_loops-hard-q8

SamamthaH:
## PR Checklist
- [ ]  EOF newlines
---
## Website

### DIFF
--- a/src/loops/nested_loops/hard/q8/MainTest.java
+++ b/src/loops/nested_loops/hard/q8/MainTest.java
@@ -29,7 +29,7 @@ public class MainTest extends BaseTest {
     public void runMain(){PrimeOrdial.main(new String[0]);}
 
     static Stream<int[]> inputProvider(){
-        return Stream.of(new int[]{4, 10, 8, 100, 20, 7}, new int[]{6, 8, 1087}, new int[]{851, 4, 23}, new int[]{-5, -8, -7, -110, 3}, new int[]{0, 3});
+        return Stream.of(new int[]{4, 10, 8, 100, 20, 7}, new int[]{6, 8, 1087}, new int[]{851, 4, 23}, new int[]{-5, -8, -7, -110, 3}, new int[]{0, 3}, new int[]{3});
     }
 
     @ParameterizedTest
--- a/src/loops/nested_loops/hard/q8/MainTest.java
+++ b/src/loops/nested_loops/hard/q8/MainTest.java
@@ -1,0 +1,57 @@
+package loops.nested_loops.hard.q8;
+
+import global.BaseTest;
+import global.exceptions.InvalidClauseException;
+import global.tools.TestOption;
+import global.variables.Clause;
+import global.variables.clauses.NewLine;
+import global.variables.clauses.PlaceHolder;
+import global.variables.clauses.StringLiteral;
+import global.variables.wrappers.Optional;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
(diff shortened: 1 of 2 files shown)


src/loops/nested_loops/hard/q8/PrimeOrdial.java
```diff
@@ -0,0 +1,37 @@
+package loops.nested_loops.hard.q8;
+
+/*
+Write a program called PrimeOrdial that will keep asking users for a number until they
+enter a prime number, after which it will break from the loop.
```
Hedgemon4, line 5: Typically, prime numbers are positive, so I think we should exclude negatives from this question and instead have an input check if the number is less than 0.

src/loops/nested_loops/hard/q8/MainTest.java
```diff
@@ -0,0 +1,57 @@
...
+                new Optional(new StringLiteral(" ")),
+                new NewLine(),
+                new PlaceHolder()
+        };
+    }
+
+    public void runMain(){PrimeOrdial.main(new String[0]);}
+
+    static Stream<int[]> inputProvider(){
+        return Stream.of(new int[]{4, 10, 8, 100, 20, 7}, new int[]{6, 8, 1087}, new int[]{851, 4, 23}, new int[]{-5, -8, -7, -110, 3}, new int[]{0, 3}, new int[]{3});
```
Hedgemon4, line 32: See my comment about negatives, and I think we should also add 1 to the one test because depending on how they code it, it could potentially pass and shouldn't
```suggestion
        return Stream.of(new int[]{4, 10, 8, 100, 20, 7}, new int[]{6, 8, 1087}, new int[]{851, 4, 23}, new int[]{-5, -8, -7, -110, 3}, new int[]{0, 1, 3}, new int[]{3});
```

Hedgemon4 (changes requested):

src/loops/nested_loops/hard/q8/PrimeOrdial.java
SamamthaH: There's no need to. Negative numbers cannot by definition be prime, and the program recognizes that. Negative numbers aren't prime and if you feed the program any negative number it will tell you it isn't prime. That seems fine to me.

Hedgemon4: Ah ok. I was not certain if the code would recognize that a negative is not prime, but if it works then that is fine.

src/loops/nested_loops/hard/q8/MainTest.java
```diff
@@ -0,0 +1,57 @@
...
+    public void runMain(){PrimeOrdial.main(new String[0]);}
+
+    static Stream<int[]> inputProvider(){
+        return Stream.of(new int[]{4, 10, 8, 100, 20, 7}, new int[]{6, 8, 1, 1087}, new int[]{851, 4, 23}, new int[]{-5, -8, -7, -110, 3}, new int[]{0, 3}, new int[]{3});
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void loopsUntilItEncountersAPrime(int[] n) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not correctly loop until it reaches a prime number.";
```
Hedgemon4, line 38:
```suggestion
        TestOption.incorrectStructureErrorMessage = "Your program does not correctly loop until it recieves a prime number as input.";
```

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #326, opened 2022-07-18
loops-simple_programs_with_repitition-easy-q2

SamamthaH:
## PR Checklist
- [ ]  EOF newlines
---
## Website

### DIFF
--- a/src/loops/simple_programs_with_repitition/easy/q2/MainTest.java
+++ b/src/loops/simple_programs_with_repitition/easy/q2/MainTest.java
@@ -1,0 +1,55 @@
+package loops.simple_programs_with_repitition.easy.q2;
+
+import global.BaseTest;
+import global.exceptions.InvalidClauseException;
+import global.tools.TestOption;
+import global.variables.Clause;
+import global.variables.clauses.NewLine;
+import global.variables.clauses.PlaceHolder;
+import global.variables.clauses.StringLiteral;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
+
+public class MainTest extends BaseTest {
+    // Parsons
+    public Clause[] testSentence(){
+        TestOption.isInputTest = true;
+        TestOption.defaultInput = "2";
+        return new Clause[]{
+                new StringLiteral("Enter either 1, 2, 3, 4: "),
+                new NewLine(),
+                new PlaceHolder()
+        };
+    }
+
+    public void runMain(){VirtualBouncer.main(new String[0]);}
+
+    static Stream<int[]> inputProvider(){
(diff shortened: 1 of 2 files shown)


src/loops/simple_programs_with_repitition/easy/q2/VirtualBouncer.java
```diff
@@ -0,0 +1,24 @@
+package loops.simple_programs_with_repitition.easy.q2;
+
+/*
+Write a program that asks the user to enter either 1, 2, 3 , or 4, and keep asking them to enter a valid value
+until they enter one of these 4 values.
+ */
```
Hedgemon4, line 6: I think this should have a sample output just to show that the message is printed multiple times (I know the question says this, but I think it better to include it than not).

src/loops/simple_programs_with_repitition/easy/q2/MainTest.java
```diff
@@ -0,0 +1,55 @@
...
+    public void runMain(){VirtualBouncer.main(new String[0]);}
+
+    static Stream<int[]> inputProvider(){
+        return Stream.of(new int[]{5, 8, 10, 11, 6, 23, 1}, new int[]{-1, -2, -3, 2}, new int[]{2});
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void loopsUntilSpecifiedNumbersAreEntered(int[] n) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not loop until 1, 2, 3, or 4 is entered.";
```
Hedgemon4, line 36:
```suggestion
        TestOption.incorrectStructureErrorMessage = "Your program does not loop until a valid number is entered.";
```

Hedgemon4 (changes requested):

src/loops/simple_programs_with_repitition/easy/q2/MainTest.java
```diff
@@ -0,0 +1,55 @@
...
+    public void runMain(){VirtualBouncer.main(new String[0]);}
+
+    static Stream<int[]> inputProvider(){
+        return Stream.of(new int[]{5, 8, 10, 11, 6, 23, 1}, new int[]{-1, -2, -3, 2}, new int[]{2});
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void loopsUntilSpecifiedNumbersAreEntered(int[] n) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not loop until a valid is entered.";
```
Hedgemon4, line 36:
```suggestion
        TestOption.incorrectStructureErrorMessage = "Your program does not loop until a valid number is entered.";
```

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #334, opened 2022-07-27
Fixed loops-programs_with_repetition-medium-q4/5 

Hedgemon4:
Formated code and added the error message which should be printed when invalid input is encountered.


src/loops/programs_with_repetition/medium/q4/OurGlass.java
```diff
@@ -2,8 +2,8 @@
 
 /*
  * Write a program to print an hourglass using the char “*” with nested loops and a positive integer n representing the width of the hourglass.
- * If the user enters a negative integer or 0, the program should print an error message.
- * 
+ * If the user enters a negative integer or 0, the program should print the error message "Invalid Input!".
```
Hedgemon4, line 5:
```suggestion
 * If the user enters a negative integer or 0, the program should print the error message "Invalid input!".
```

PR #345, opened 2022-08-03
methods-using_methods-hard-q4

SamamthaH:
## PR Checklist
- [ ]  EOF newlines
---
## Website

### DIFF
--- a/src/methods/using_methods/hard/q4/HowOld.java
+++ b/src/methods/using_methods/hard/q4/HowOld.java
@@ -1,7 +1,7 @@
 package methods.using_methods.hard.q4;
 
 /*
-Write a Java Program called Using4 that asks users for the current year and for the year that they were born in.
+Write a Java Program called HowOld that asks users for the current year and for the year that they were born in.
 Then, write a method called ageCalc that accepts both values as input parameters, checks that both years are positive
  and that the current year is greater than or equal to the year the user was born and returns the age of the user.
  If not, return -1.
--- a/src/methods/using_methods/hard/q4/HowOld.java
+++ b/src/methods/using_methods/hard/q4/HowOld.java
@@ -1,10 +1,10 @@
 package methods.using_methods.hard.q4;
 
 /*
-Write a Java Program called HowOld that asks users for the current year and for the year that they were born in.
-Then, write a method called ageCalc that accepts both values as input parameters, checks that both years are positive
- and that the current year is greater than or equal to the year the user was born and returns the age of the user.
- If not, return -1.
+Write a Java Program called HowOld that asks users for the current year and the year they were born.
+Then, write a method called ageCalc that accepts both values as input parameters and calculates and returns the user's age.
+This method should check that both years are positive and that the current year is greater than or equal to the year
+the user was born. If not, return -1.
 
 Sample output:
 Please enter the current year:
--- a/src/methods/using_methods/hard/q4/HowOld.java
+++ b/src/methods/using_methods/hard/q4/HowOld.java
@@ -2,7 +2,9 @@ package methods.using_methods.hard.q4;
 
 /*
 Write a Java Program called Using4 that asks users for the current year and for the year that they were born in.
-Then, write a method called ageCalc that accepts both values as input parameters and returns the age of the user.
+Then, write a method called ageCalc that accepts both values as input parameters, checks that both years are positive
+ and that the current year is greater than or equal to the year the user was born and returns the age of the user.
+ If not, return -1.
 
 Sample output:
 Please enter the current year:
@@ -13,7 +15,7 @@ Your age is 24
  */
 
 import java.util.Scanner;
-public class Using4
+public class HowOld
 {
     public static void main( String[] args )
     {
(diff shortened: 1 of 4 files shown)


src/methods/using_methods/hard/q4/MainTest.java
```diff
@@ -0,0 +1,57 @@
...
+                new StringLiteral("Please enter the current year:"),
+                new NewLine(),
+                new StringLiteral("Please enter the year you were born in:"),
+                new NewLine(),
+                new StringLiteral("Your age is "),
+                new IntegerLiteral("age")
+        };
+    }
+
+    public void runMain(){Using4.main(new String[0]);}
```
Hedgemon4, line 34: See other comment about changing the class name

src/methods/using_methods/hard/q4/Using4.java
```diff
@@ -0,0 +1,31 @@
...
+        Scanner input = new Scanner(System.in);
+        System.out.println("Please enter the current year:");
+        int currYear = input.nextInt();
+        System.out.println("Please enter the year you were born in:");
+        int yearBorn = input.nextInt();
+        System.out.println("Your age is " + ageCalc(currYear,yearBorn));
+    }
+    public static int ageCalc(int a, int b)
+    {
+        return a-b;
```
Hedgemon4, line 29: I think we should maybe change this to make sure both the years are positive, and that the current year is greater than or equal to the year born

Hedgemon4: Just because it doesn't make sense to return a negative age

Hedgemon4 (changes requested):

src/methods/using_methods/hard/q4/HowOld.java
```diff
@@ -0,0 +1,38 @@
+package methods.using_methods.hard.q4;
+
+/*
+Write a Java Program called HowOld that asks users for the current year and for the year that they were born in.
```
Hedgemon4, line 4: Change question text to:
`
Write a Java Program called HowOld that asks users for the current year and the year they were born. Then, write a method called ageCalc that accepts both values as input parameters and calculates and returns the user's age. This method should check that both years are positive and that the current year is greater than or equal to the year the user was born. If not, return -1.
`

src/methods/using_methods/hard/q4/MainTest.java
```diff
@@ -0,0 +1,68 @@
...
+    void correctAgeCalcInvalidInputMethod(int tooOld, int tooNew) throws Throwable {
+        Object output = MethodUtil.invokeIfMethodExists(HowOld.class, "ageCalc", new Object[]{tooOld, tooNew}, int.class, int.class);
+        CustomAssertions._assertEquals(-1, output, "Your ageCalc method does not correctly identify invalid input.");
+    }
+
+    @ParameterizedTest
+    @MethodSource("mainMethodInputProvider")
+    void printsOutputCorrectly(String in, int age){
+        runWithInput(in);
+        assertEquals(String.valueOf(age), getItemByName("age"), "Your program does not display the correct age.");
```
Hedgemon4, line 66: This is picky but I think this should be done as an integer assert and not a string assert

src/methods/using_methods/hard/q4/MainTest.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public void runMain(){HowOld.main(new String[0]);}
+
+    static Stream<Arguments> ageCalcInputProvider(){
+        return Stream.of(Arguments.of(9, 0, 9), Arguments.of(2031, 2000, 31), Arguments.of(1734, 1562, 172));
+    }
+
+    static Stream<Arguments> ageCalcInvalidInputProvider(){
+        return Stream.of(Arguments.of(2019, 2020), Arguments.of(2017, 2020), Arguments.of(1999, 2078));
```
Hedgemon4, line 41: This doesn't have a test case for negative input

Hedgemon4 (changes requested):

src/methods/using_methods/hard/q4/MainTest.java
```diff
@@ -0,0 +1,68 @@
...
+    @ParameterizedTest
+    @MethodSource("ageCalcInputProvider")
+    void correctAgeCalcMethod(int curr, int old, int age) throws Throwable {
+        Object output = MethodUtil.invokeIfMethodExists(HowOld.class, "ageCalc", new Object[]{curr, old}, int.class, int.class);
+        CustomAssertions._assertEquals(age, output, "Your ageCalc method does not properly calculate age.");
+    }
+
+    @ParameterizedTest
+    @MethodSource("ageCalcInvalidInputProvider")
+    void correctAgeCalcInvalidInputMethod(int tooOld, int tooNew) throws Throwable {
```
Hedgemon4, line 57: I think this name should be changed to something more clear.

Hedgemon4:
```suggestion
    void ageCalcIdentifiesInvalidInputCorrectly(int tooOld, int tooNew) throws Throwable {
```

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #353, opened 2022-08-05
methods-using_methods-hard-q9

SamamthaH:
## PR Checklist
- [ ]  EOF newlines
---
## Website

### DIFF
--- a/src/methods/using_methods/hard/q9/MainTest.java
+++ b/src/methods/using_methods/hard/q9/MainTest.java
@@ -1,4 +1,55 @@
 package methods.using_methods.hard.q9;
 
-public class MainTest {
+import global.BaseTest;
+import global.tools.CustomAssertions;
+import global.tools.TestOption;
+import global.utils.MethodUtil;
+import global.variables.Clause;
+import global.variables.clauses.DoubleLiteral;
+import global.variables.clauses.NewLine;
+import global.variables.clauses.StringLiteral;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MainTest extends BaseTest {
+    // Parsons
+    public Clause[] testSentence() {
+        TestOption.isInputTest = true;
+        TestOption.defaultInput = "7";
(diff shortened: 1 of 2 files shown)


src/methods/using_methods/hard/q9/MainTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MainTest extends BaseTest {
+    // Parsons
+    public Clause[] testSentence() {
+        TestOption.isInputTest = true;
+        TestOption.defaultInput = "7";
+        return new Clause[]{
+                new StringLiteral("How many people are going on the trip:"),
```
Hedgemon4, line 25:
```suggestion
                new StringLiteral("How many people are going on the trip?"),
```

src/methods/using_methods/hard/q9/Tickets.java
```diff
@@ -0,0 +1,35 @@
...
+total cost of the tickets.
+ */
+
+import java.util.Scanner;
+public class Tickets
+{
+    public static void main( String[] args )
+    {
+        Scanner input = new Scanner(System.in);
+        System.out.println("How many people are going on the trip:");
```
Hedgemon4, line 18:
```suggestion
        System.out.println("How many people are going on the trip?");
```

src/methods/using_methods/hard/q9/Tickets.java
```diff
@@ -0,0 +1,35 @@
+package methods.using_methods.hard.q9;
+
+/*
+Write a Java Program called Tickets that asks users the number of people getting a ticket for a trip.
```
Hedgemon4, line 4: Change question text to
`
Write a Java Program called Tickets that asks users the number of people getting a ticket for a trip. Your program should have a method called tripCostCalc that accepts the number of people as an input parameter and returns the total cost of the tickets. For two or fewer, the cost of tickets is $5.00 per person; for 3-5 people, the price is $4.50 per person; for 6-10 people, the price is $4.00 per person; for 11 or more people, the price is $3.00 per person. Finally, print the total cost of the trip. 
`

Hedgemon4: We should also maybe add to print 0 if the number of tickets entered is less than 0

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #369, opened 2022-08-12
methods-defining_methods-hard-q3

ketphan02:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/methods/defining_methods/hard/q3/Polygo.java
```diff
@@ -0,0 +1,29 @@
...
+public class Polygo {
+  public static void main(String[] args) {
+    Scanner input = new Scanner(System.in);
+    System.out.println("Enter the number of sides on the polygon:");
+    int countSides = input.nextInt();
+    System.out.println("The sum of interior angles of the polygon is: " + innerAngles(countSides));
+
+  }
+
+  public static int innerAngles(int a) {
```
Hedgemon4, line 25: I think we should change the questions so that the innerAngles should check to make sure the passed integer is greater than or equal to 3, and return -1 if it is not

src/methods/defining_methods/hard/q3/Polygo.java
```diff
@@ -0,0 +1,29 @@
+package methods.defining_methods.hard.q3;
+
+/**
+ * [J] Write a Java Program that asks the user for the number of sides on a polygon, then determines the sum of the measure
```
Hedgemon4, line 4: Change question text to: 
`Write a Java Program that asks the user for the number of sides on a polygon, then determines the sum of the measure of the interior angles of that polygon. Remember that the formula to determine the sum of interior angles is (n-2)*180, where n is the number of sides. Create a method called "innerAngles(int size)" that takes the number of sides and returns an integer as the sum of the interior angles. The input is guaranteed to be a positive integer greater or equal to 3.`

src/methods/defining_methods/hard/q3/MainTest.java
```diff
@@ -0,0 +1,64 @@
...
+      Arguments.of(3, 180),
+      Arguments.of(100, 17640),
+      Arguments.of(10000, 1799640)
+    );
+  }
+
+  @ParameterizedTest
+  @MethodSource("innerAnglesInputProvider")
+  void correctInnerAnglesMethod(int a, int innerAngles) throws Throwable {
+    String errorMessage = "Your \"innerAngles()\" method does not correctly calculated the inner angles of a polygon with " + a + " sides.";
```
Hedgemon4, line 49: We should not tell them this a value because as this is a java question, they could just hardcode the values if they failed enough times

src/methods/defining_methods/hard/q3/MainTest.java
```diff
@@ -0,0 +1,64 @@
...
+    Object output = MethodUtil.invokeIfMethodExists(Polygo.class, "innerAngles", new Object[]{a}, int.class);
+    CustomAssertions._assertEquals(innerAngles, output, errorMessage);
+  }
+
+  @ParameterizedTest
+  @MethodSource("innerAnglesInputProvider")
+  void printsOutputCorrectly(int a, int innerAngles) throws InvalidClauseException {
+    TestOption.incorrectStructureErrorMessage = "Your program does not correctly calculate and print the inner angles of a polygon with " + a + " sides.";
```
Hedgemon4, line 57: Same as the other error message

Hedgemon4 (changes requested):

src/methods/defining_methods/hard/q3/Polygo.java
SamamthaH: since the question text says assume input is 3 or higher I'm going to say its fine as is

src/methods/defining_methods/hard/q3/Polygo.java
```diff
@@ -0,0 +1,31 @@
+package methods.defining_methods.hard.q3;
+
+/*
+Write a Java Program that asks the user for the number of sides on a polygon, then determines the sum of the measure of
+the interior angles of that polygon. Remember that the formula to determine the sum of interior angles is (n-2)*180,
+where n is the number of sides. Create a method called "innerAngles(int size)" that takes the number of sides and
```
Hedgemon4, line 6: I think that because this is a hard question in one of the later categories, we can add in that the method should check that the input is greater than 2, and return -1 otherwise

src/methods/defining_methods/hard/q3/MainTest.java
```diff
@@ -0,0 +1,69 @@
...
+        MethodTest m = new MethodTest(Polygo.class, "innerAngles", arguments);
+        Object output = m.callMethod();
+        String errorMessage = "Your innerAngles() method does not correctly calculated the inner angles of a polygon.";
+        CustomAssertions._assertEquals(innerAngles, output, errorMessage);
+    }
+
+    @ParameterizedTest
+    @MethodSource("innerAnglesInputProvider")
+    void printsOutputCorrectly(int a, int innerAngles) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not correctly calculate and print the inner angles of a polygon.";
```
Hedgemon4, line 63: Should probably say something about it is in degrees

Hedgemon4: Like sum of the inner angles in degrees (you can word it how you see fit)

src/methods/defining_methods/hard/q3/MainTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @ParameterizedTest
+    @MethodSource("innerAnglesInputProvider")
+    void correctInnerAnglesMethod(int a, int innerAngles) throws Throwable {
+        Object[][] arguments = {
+                {a, int.class}
+        };
+        MethodTest m = new MethodTest(Polygo.class, "innerAngles", arguments);
+        Object output = m.callMethod();
+        String errorMessage = "Your innerAngles() method does not correctly calculated the inner angles of a polygon.";
```
Hedgemon4, line 56: Should be calculate

Hedgemon4: Also, say something about the sum of the inner angles in degrees (so they know not to use radians)

Hedgemon4 (changes requested):

src/methods/defining_methods/hard/q3/Polygo.java
```diff
@@ -0,0 +1,34 @@
...
+returns an integer as the sum of the interior angles. If the passed integer is less than 3, return -1..
```
Hedgemon4, line 7:
```suggestion
returns an integer as the sum of the interior angles. If the passed integer is less than 3, return -1.
```

src/methods/defining_methods/hard/q3/MainTest.java
```diff
@@ -0,0 +1,72 @@
...
+public class MainTest extends BaseTest {
+    // Java
+
+    public Clause[] testSentence() {
+        TestOption.isInputTest = true;
+        TestOption.defaultInput = "8";
+
+        return new Clause[]{
+                new StringLiteral("Enter the number of sides on the polygon:"),
+                new NewLine(),
```
Hedgemon4, line 30: Add an optional to the end of the line for an empty space

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #372, opened 2022-08-12
methods-modular_programs-easy-q6

SamamthaH:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/methods/modular_programs/easy/q6/Century.java
```diff
@@ -0,0 +1,31 @@
+package methods.modular_programs.easy.q6;
+
+/*
+Write a Java Program that asks users for the year that they were born in, then returns what century they were born in.
```
Hedgemon4, line 4: Change question text to:
`
Write a Java Program that asks users for the year they were born, then returns what century that year is in. The program should have a method called centuryCalc to determine this, which works for any century from the 1500's to 2000's. If the input is in any other century, return "unknown century" .
`

src/methods/modular_programs/easy/q6/Century.java
```diff
@@ -0,0 +1,31 @@
...
+public class Century
+{
+    public static void main( String[] args )
+    {
+        Scanner input = new Scanner(System.in);
+        System.out.println("Please enter the year you were born in: ");
+        int userYear = input.nextInt();
+        System.out.println("You're from the " + centuryCalc(userYear));
+    }
+    public static String centuryCalc(int userYear)
```
Hedgemon4, line 22: I think we should update this question to work with more centuries (see question text).

Hedgemon4 (changes requested):

src/methods/modular_programs/easy/q6/Century.java
```diff
@@ -0,0 +1,40 @@
+package methods.modular_programs.easy.q6;
+
+/*
+Write a Java Program that asks users for the year they were born, then returns what century that year is in. The program
```
Hedgemon4, line 4: Change the question text to:
`Write a Java Program that asks users for the year they were born, then prints the century the individual was born. The program should have a method called centuryCalc, which determines the century and works for any century from the 1500s to the 2000's. If the input is in any other century, return "Unknown Century".
`

Hedgemon4: We should also maybe add a note that the 1500s is the 16th century and so on

src/methods/modular_programs/easy/q6/MainTest.java
```diff
@@ -0,0 +1,88 @@
...
+                Arguments.of("1634", "17th Century"),
+                Arguments.of("1555", "16th Century"),
+                Arguments.of("1265", "Unknown Century"));
+    }
+
+    @ParameterizedTest
+    @MethodSource("centuryCalcInputProvider")
+    void correctCenturyCalcMethod(int in, String century) throws Throwable {
+        Object output = MethodUtil.invokeIfMethodExists(Century.class, "centuryCalc", new Object[]{in}, int.class);
+        CustomAssertions._assertEquals(century, output, "Your centuryCalc method does not return the right century.");
```
Hedgemon4, line 77:
```suggestion
        CustomAssertions._assertEquals(century, output, "Your centuryCalc method does not return the correct century.");
```

src/methods/modular_programs/easy/q6/MainTest.java
```diff
@@ -0,0 +1,88 @@
...
+                Arguments.of("1782", "18th Century"),
+                Arguments.of("1634", "17th Century"),
+                Arguments.of("1555", "16th Century"),
+                Arguments.of("1265", "Unknown Century"));
+    }
+
+    @ParameterizedTest
+    @MethodSource("centuryCalcInputProvider")
+    void correctCenturyCalcMethod(int in, String century) throws Throwable {
+        Object output = MethodUtil.invokeIfMethodExists(Century.class, "centuryCalc", new Object[]{in}, int.class);
```
Hedgemon4, line 76: Change to methodTest

Hedgemon4 (changes requested):

src/methods/modular_programs/easy/q6/MainTest.java
```diff
@@ -0,0 +1,92 @@
...
+                {in, int.class}
+        };
+        MethodTest m = new MethodTest(Century.class, "centuryCalc", arguments);
+        Object output = m.callMethod();
+        CustomAssertions._assertEquals(century, output, "Your centuryCalc method does not return the correct century.");
+    }
+
+    @ParameterizedTest
+    @MethodSource("mainMethodInputProvider")
+    void printsOutputCorrectly(String in, String century) throws InvalidClauseException {
```
Hedgemon4, line 86: Again, if we are providing an int as input, I think we should provide the method with int and then String.valueOf(x) in the method

src/methods/modular_programs/easy/q6/MainTest.java
```diff
@@ -0,0 +1,92 @@
...
+                Arguments.of(1500, "16th Century"),
+                Arguments.of(1499, "Unknown Century"),
+                Arguments.of(1498, "Unknown Century"),
+                Arguments.of(567, "Unknown Century"),
+                Arguments.of(-5, "Unknown Century"));
+    }
+
+    static Stream<Arguments> mainMethodInputProvider() {
+        return Stream.of(
+                Arguments.of("2022", "21st Century"),
```
Hedgemon4, line 64: Change to provide int instead.

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #374, opened 2022-08-13
methods-method_overloading-hard-q8

ketphan02:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/methods/method_overloading/hard/q8/Overload8.java
```diff
@@ -0,0 +1,28 @@
...
+ * with the data type of the number that was passed in into the method. Use
+ * method overloading to only call one method and let's call that method
+ * printType.
+ * 
+ * Sample Output:
+ *  The data type associated with 14 is : int
+ *  The data type associated with 61.34 is : double
+ */
+
+public class Overload8 {
```
Hedgemon4, line 14: I think we should change the class name to something to do with the question

src/methods/method_overloading/hard/q8/MainTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+import global.BaseTest;
+import global.tools.CustomAssertions;
+import global.utils.MethodUtil;
+import global.variables.Clause;
+import global.variables.clauses.NewLine;
+import global.variables.clauses.StringLiteral;
+
+public class MainTest extends BaseTest {
+  // Java
```
Hedgemon4, line 21: As this question is a Java question, we have to change the question so that students can not hardcode the output. Due to how the scanner works, we will probably have to have them ask for an integer and a double separately, and then call the printType method separately.

Hedgemon4: Also, did you change the original question so that the printType method returns a string instead of printing the type in that method? Because we can check printed output in a method.

Hedgemon4 (changes requested):

src/methods/method_overloading/hard/q8/FindDataType.java
```diff
@@ -0,0 +1,37 @@
+package methods.method_overloading.hard.q8;
+
+/*
+Write a Java Program that asks the user for an integer then a double and prints out a message with the data type of the
```
Hedgemon4, line 4: Change question text to:
`Write a Java Program that asks the user for an integer and a double and then prints out a message with the data type of each number. To identify the data type, write two methods called printType, one that accepts an int and returns the string "int" and one that takes a double and returns the string "double". `

src/methods/method_overloading/hard/q8/MainTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @ParameterizedTest
+    @MethodSource("inputIntegerProvider")
+    void correctPrintTypeIntegerMethod(int intNumber) throws Throwable {
+        Object[][] arguments = {
+                {intNumber, int.class}
+        };
+        MethodTest m = new MethodTest(FindDataType.class, "printType", arguments);
+        Object output = m.callMethod();
+        String errorMessage = "Your printType() method for integer does not print the correct type.";
```
Hedgemon4, line 66:
```suggestion
        String errorMessage = "Your printType() method for integers does not return the correct string.";
```

src/methods/method_overloading/hard/q8/MainTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @ParameterizedTest
+    @MethodSource("inputDoubleProvider")
+    void correctPrintTypeDoubleMethod(double doubleNumber) throws Throwable {
+        Object[][] arguments = {
+                {doubleNumber, double.class}
+        };
+        MethodTest m = new MethodTest(FindDataType.class, "printType", arguments);
+        Object output = m.callMethod();
+        String errorMessage = "Your printType() method for double does not print the correct type.";
```
Hedgemon4, line 78:
```suggestion
        String errorMessage = "Your printType() method for doubles does not return the correct type.";
```

src/methods/method_overloading/hard/q8/MainTest.java
```diff
@@ -0,0 +1,95 @@
...
+                Arguments.of(5, 3.54),
+                Arguments.of(-67, -438462.56),
+                Arguments.of(789317, -23.4316),
+                Arguments.of(-4, 45.21)
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputIntegerProvider")
+    void correctPrintTypeIntegerMethod(int intNumber) throws Throwable {
```
Hedgemon4, line 60: I think to be more consistent with our usual format, we should call this "correctPrintTypeMethodForIntegers" instead, and similarly for the double one.

Hedgemon4: Unless you think that is too confusing for the students?

src/methods/method_overloading/hard/q8/MainTest.java
```diff
@@ -0,0 +1,95 @@
...
+        String errorMessage = "Your printType() method for double does not print the correct type.";
+        CustomAssertions._assertEquals(output, "double", errorMessage);
+    }
+
+    @ParameterizedTest
+    @MethodSource("mainMethodInput")
+    void printsCorrectOutput(int x, double y) throws InvalidClauseException {
+        runWithInput(x + " " + y, new Clause[][]{
+                {
+                        new StringLiteral("The data type associated with " + x + " is: "),
```
Hedgemon4, line 87: Does all of this have to be in the placeholder? Can't we have two separate placeholders?

src/methods/method_overloading/hard/q8/MainTest.java
```diff
@@ -0,0 +1,95 @@
...
+        };
+        MethodTest m = new MethodTest(FindDataType.class, "printType", arguments);
+        Object output = m.callMethod();
+        String errorMessage = "Your printType() method for double does not print the correct type.";
+        CustomAssertions._assertEquals(output, "double", errorMessage);
+    }
+
+    @ParameterizedTest
+    @MethodSource("mainMethodInput")
+    void printsCorrectOutput(int x, double y) throws InvalidClauseException {
```
Hedgemon4, line 84: We should also probably change the error message

Hedgemon4 (changes requested):

src/methods/method_overloading/hard/q8/MainTest.java
SamamthaH: isn't the point of placeholders that they capture all the input left because they can't be sure when you need them to stop?

Hedgemon4: The rest should be static

Hedgemon4 (changes requested):
See my discord message

src/methods/method_overloading/hard/q8/FindDataType.java
```diff
@@ -0,0 +1,29 @@
+package methods.method_overloading.hard.q8;
+
+/*
+Write a Java Program with an overloaded method called printType which can take either an integer or a double. In the
+given main method, printType is passed the integer 14 and on a new line the double 61.34. You must write the methods and
```
Hedgemon4, line 5:
```suggestion
given main method, printType is passed the integer 14 and on a new line the double 61.34. You must write the overloaded methods and
```

Hedgemon4 (approved):

PR #376, opened 2022-08-13
methods-defining_methods-hard-q7

ketphan02:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/methods/defining_methods/hard/q7/CurrentYearMethod.java
```diff
@@ -0,0 +1,31 @@
+package methods.defining_methods.hard.q7;
+
+/**
+ * [J] Write a Java Program that asks users for the number of days in the current year. Then write a
+ * method that accepts that value and returns a boolean value if it is a leap year or not (true if it
+ * is a leap year, false if not). Let’s call the method “leapChecker”.
```
Hedgemon4, line 6: Change question text to: 
`Write a Java Program that asks users for the number of days in the current year. Then write a method called leapChecker that accepts that value as an integer and returns a boolean value if it is a leap year or not (true if it is, false if not). The year is a leap year if it has 366 days. Lastly, if the method returns true, print "This year is a leap year", if it is false, print "This year isn't a leap year". `

src/methods/defining_methods/hard/q7/MainTest.java
```diff
@@ -0,0 +1,65 @@
...
+    };
+  }
+
+  public void runMain() {
+    CurrentYearMethod.main(new String[0]);
+  }
+
+  static Stream<Arguments> possibleNumDaysForMethod() {
+    return Stream.of(
+        Arguments.of(366, true),
```
Hedgemon4, line 38: For this question, we should also specify what happens if an invalid number is entered (should probably just return false)

Hedgemon4: Then we also need to add test cases for this

Hedgemon4 (changes requested):

src/methods/defining_methods/hard/q7/MainTest.java
SamamthaH: I don't know, it's only to check if the year is a leap year, so any invalid input would still be false

Hedgemon4: Fair enough

src/methods/defining_methods/hard/q7/MainTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+    static Stream<Arguments> possibleNumDaysForMethod() {
+        return Stream.of(
+                Arguments.of(366, true),
+                Arguments.of(365, false),
+                Arguments.of(34, false),
+                Arguments.of(367, false));
+    }
+
+    static Stream<Arguments> possibleNumDaysForOutput() {
```
Hedgemon4, line 45: We should test negatives though because we did not day to assume valid input

Hedgemon4: We should also probably have a few more test cases (just in case).

src/methods/defining_methods/hard/q7/MainTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @ParameterizedTest
+    @MethodSource("possibleNumDaysForMethod")
+    void correctLeapCheckerMethod(int numDays, boolean isLeapYear) throws Throwable {
+        Object[][] arguments = {
+                {numDays, int.class}
+        };
+        MethodTest m = new MethodTest(CurrentYearMethod.class, "leapChecker", arguments);
+        Object output = m.callMethod();
+        String errorMessage = "Your leapChecker() method does not return the correct result";
```
Hedgemon4, line 61: Change the error message to be grammatically correct.

src/methods/defining_methods/hard/q7/MainTest.java
```diff
@@ -0,0 +1,73 @@
...
+        CustomAssertions._assertEquals(isLeapYear, output, errorMessage);
+    }
+
+    @ParameterizedTest
+    @MethodSource("possibleNumDaysForOutput")
+    void printsCorrectLeapYearMessage(int numDays, String isLeapYearStr) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not print the correct message for this year";
```
Hedgemon4, line 68: Change error message to be grammatically correct

Hedgemon4 (changes requested):

src/methods/defining_methods/hard/q7/MainTest.java
```diff
@@ -0,0 +1,81 @@
...
+        String errorMessage = "Your leapChecker() method does not return the correct result.";
+        CustomAssertions._assertEquals(isLeapYear, output, errorMessage);
+    }
+
+    @ParameterizedTest
+    @MethodSource("possibleNumDaysForOutput")
+    void printsCorrectLeapYearMessage(int numDays, String isLeapYearStr) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not print the correct message for this year.";
+        runWithInput(numDays + "", new Clause[]{
+                new StringLiteral(isLeapYearStr),
```
Hedgemon4, line 78: Can we add an optional to the end of the placeholder if they print an extra space?

Hedgemon4 (changes requested):

Hedgemon4 (changes requested):
See my discord message

Hedgemon4 (approved):

PR #378, opened 2022-08-13
methods-defining_methods-hard-q5

ketphan02:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/methods/defining_methods/hard/q5/FavMovie.java
```diff
@@ -0,0 +1,39 @@
+package methods.defining_methods.hard.q5;
+
+/**
+ * [J] Write a Java Program that asks users for their favourite movie and asks them how many times they've watched that movie. Then,
```
Hedgemon4, line 4: Change question text to:
`Write a Java program asking users for their favourite movie and how many times they've watched that movie. Then, write a Java Method that takes the string and integer as parameters and returns a string. The returned string should be as follows: if the integer is less than or equal to five, then return a string saying: "You've watched <the movie name> a few times!"; if the integer is less than or equal to 10 and greater than 5, return a string saying: "You've watched <the movie name> a lot of times!"; if the integer is greater than 10, then return a string saying: "You've watched <the movie name> so many times! Wow!". Call your method movieCounter and pass in the movie name and the number of times you have watched the movie respectively.`

src/methods/defining_methods/hard/q5/MainTest.java
```diff
@@ -0,0 +1,68 @@
...
+        new PlaceHolder(),
+    };
+  }
+
+  public void runMain() {
+    FavMovie.main(new String[0]);
+  }
+
+  static Stream<Arguments> inputProvider() {
+    return Stream.of(
```
Hedgemon4, line 38: You should change the movie name in some of these

src/methods/defining_methods/hard/q5/MainTest.java
```diff
@@ -0,0 +1,68 @@
...
+    Object output = MethodUtil.invokeIfMethodExists(FavMovie.class, "movieCounter",
+        new Object[] { movieName, timeWatched }, String.class, int.class);
+    CustomAssertions._assertEquals(output, exclaimatory, errorMesasge);
+  }
+
+  @ParameterizedTest
+  @MethodSource("inputProvider")
+  void printsCorrectOutput(String movieName, int timeWatched, String exclaimatory) throws InvalidClauseException {
+    TestOption.incorrectStructureErrorMessage = "Your program does not print the correct exclaimatory message for the movie "
+        + movieName + ".";
```
Hedgemon4, line 63: I would not give the user the movie name because they could just hardcode the answer

src/methods/defining_methods/hard/q5/FavMovie.java
```diff
@@ -0,0 +1,39 @@
...
+ * write a Java Method that takes the string and integer as parameters and returns a string. The returned string should be as follows.
+ * If the integer is 5 or lower, then just return a string saying: "You've watched <the movie name> a few times!", if the integer is
+ * smaller than 10 and smaller than 5, then return a string saying: "You've watched <the movie name> a lot of times!", if the integer
+ * is greater than 10, then return a string saying: "You've watched <the movie name> so many times! Wow!". Call your method movieCounter
+ * and pass in the movie name and the number of times you have watched the movie respectively.
```
Hedgemon4, line 9: We should also probably specify what to do with negative numbers because they don't make sense in this context

Hedgemon4 (changes requested):

src/methods/defining_methods/hard/q5/MainTest.java
```diff
@@ -0,0 +1,75 @@
...
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void correctMethodMovieCounter(String movieName, int timeWatched, String exclaimatory) throws Throwable {
+        Object[][] arguments = {
+                {movieName, String.class},
+                {timeWatched, int.class}
+        };
+        MethodTest m = new MethodTest(FavMovie.class, "movieCounter", arguments);
+        Object output = m.callMethod();
+        String errorMessage = "Your movieCounter() method does not return the correct message.";
```
Hedgemon4, line 63:
```suggestion
        String errorMessage = "Your movieCounter() method does not return the correct message based on the move name and the number of times the user watched the moveie.";
```

src/methods/defining_methods/hard/q5/MainTest.java
```diff
@@ -0,0 +1,75 @@
...
+        CustomAssertions._assertEquals(output, exclaimatory, errorMessage);
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void printsCorrectOutput(String movieName, int timeWatched, String exclaimatory) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not print the correct message for the movie.";
```
Hedgemon4, line 70:
```suggestion
        TestOption.incorrectStructureErrorMessage = "Your program does not print the correct message for the input movie and the number of times the user watched it.";
```

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #379, opened 2022-08-13
methods-defining_methods-hard-q6

ketphan02:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/methods/defining_methods/hard/q6/IntegerMulter.java
```diff
@@ -0,0 +1,28 @@
+package methods.defining_methods.hard.q6;
+
+/**
+ * *[J] Write a Java Program that asks users for an integer between 0 and 10, then write a method
```
Hedgemon4, line 4: Change question text to: 
`Write a Java program that asks users for an integer between 0 and 10 (inclusive), which has a method called multiply1000 that takes an integer as an input parameter and returns that integer multiplied by 1000. If the method receives an invalid integer ( less than 0 or greater than 10), return -1. Call the method with the user input and print the returned integer in the main method. `

src/methods/defining_methods/hard/q6/MainTest.java
```diff
@@ -0,0 +1,70 @@
...
+        new PlaceHolder(),
+    };
+  }
+
+  public void runMain() {
+    IntegerMulter.main(new String[0]);
+  }
+
+  static Stream<Arguments> numberBetween0And10() {
+    return Stream.of(
```
Hedgemon4, line 38: Add cases for invalid numbers

Hedgemon4 (changes requested):

src/methods/defining_methods/hard/q6/MainTest.java
```diff
@@ -0,0 +1,70 @@
...
+        Arguments.of(6, 6000),
+        Arguments.of(7, 7000),
+        Arguments.of(8, 8000),
+        Arguments.of(9, 9000),
+        Arguments.of(10, 10000));
+  }
+
+  @ParameterizedTest
+  @MethodSource("numberBetween0And10")
+  void correctReturnMultiply1000(int num, int newNum) throws Throwable {
```
Hedgemon4, line 54:
```suggestion
  void correctMultiply1000Method(int num, int newNum) throws Throwable {
```

src/methods/defining_methods/hard/q6/MainTest.java
```diff
@@ -0,0 +1,70 @@
...
+    String errorMessage = "Your method \"multiply1000()\" does not return the correct new number.";
```
Hedgemon4, line 55:
```suggestion
    String errorMessage = "Your method \"multiply1000()\" does not return the correct number.";
```

Hedgemon4 (changes requested):

src/methods/defining_methods/hard/q6/MainTest.java
```diff
@@ -0,0 +1,82 @@
...
+    // Java
+
+    public Clause[] testSentence() {
+        TestOption.isInputTest = true;
+        TestOption.defaultInput = "5";
+
+        return new Clause[]{
+                new StringLiteral("Enter a number:"),
+                new NewLine(),
+                new PlaceHolder(),
```
Hedgemon4, line 30: This doesn't need a placeholder

Hedgemon4 (changes requested):

src/methods/defining_methods/hard/q6/IntegerMulter.java
```diff
@@ -0,0 +1,31 @@
+package methods.defining_methods.hard.q6;
+
+/*
+Write a Java program that asks users for an integer between 0 and 10 (inclusive), which has a method called multiply1000
```
Hedgemon4, line 4: Change question text to:
`Write a Java program that asks users for an integer between 0 and 10 (inclusive) and then multiples it by 1000. To multiply the number, create a  method called multiply1000 that takes an integer as an input parameter and returns that integer multiplied by 1000. If the method receives an invalid integer ( less than 0 or greater than 10), return -1. Call the method with the user input and print the returned integer in the main method.`

Hedgemon4 (changes requested):

Hedgemon4 (approved):

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #380, opened 2022-08-15
methods-defining_methods-hard-q9

ketphan02:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/methods/defining_methods/hard/q9/NumCoursesMethod.java
```diff
@@ -0,0 +1,26 @@
+package methods.defining_methods.hard.q9;
+
+/**
+ * [J] Write a Java Program that asks users for the number of courses they are taking right now. A course will contain
```
Hedgemon4, line 4: Write a Java program asking the user for the number of courses they are taking right now. Then, create a method named creditsCac that accepts the number of courses as an integer and returns an integer for credits the student will receive, assuming that each course consists of three credits. Print the number of credits the user will receive in the main method.

src/methods/defining_methods/hard/q9/NumCoursesMethod.java
```diff
@@ -0,0 +1,26 @@
...
+    Scanner input = new Scanner(System.in);
+    System.out.println("Enter the number of courses you're taking this year:");
+    int courses = input.nextInt();
+    System.out.println("The number of credits you'll receive is " + creditsCalc(courses));
+  }
+
+  public static int creditsCalc(int coursesCount) {
+    return 3 * coursesCount;
+  }
+}
```
Hedgemon4, line 26: Add another line to this file

src/methods/defining_methods/hard/q9/NumCoursesMethod.java
```diff
@@ -0,0 +1,26 @@
+package methods.defining_methods.hard.q9;
+
+/**
+ * [J] Write a Java Program that asks users for the number of courses they are taking right now. A course will contain
+ * 3 credits. Write a method named creditsCac that accepts the number of courses and returns the number of credits the student will receive.
+ *
```
Hedgemon4, line 6: We should also probably have a return -1 if the number of credits is less than 0

src/methods/defining_methods/hard/q9/MainTest.java
```diff
@@ -0,0 +1,63 @@
...
+        Arguments.of(4, 12),
+        Arguments.of(50, 150),
+        Arguments.of(0, 0),
+        Arguments.of(10000, 30000));
+  }
+
+  @ParameterizedTest
+  @MethodSource("coursesInputProvider")
+  void methodCreditsCalcCorrect(int courses, int credits) throws Throwable {
+    String errorMsg = "Your method \"creditsCalc()\" does not correctly calculate the number of credits";
```
Hedgemon4, line 48: Maybe add to this errorMsg just to make it more specific, like "... the number of credits the user will earn."

src/methods/defining_methods/hard/q9/MainTest.java
```diff
@@ -0,0 +1,63 @@
...
+  @ParameterizedTest
+  @MethodSource("coursesInputProvider")
+  void printsCorrectMessage(int courses, int credits) throws InvalidClauseException {
+    TestOption.incorrectStructureErrorMessage = "Your program does not print the correct credit message";
+    runWithInput(courses + "", new Clause[][] { {
+        new StringLiteral("The number of credits you'll receive is "),
+        new IntegerLiteral(credits),
+    } });
+  }
+}
```
Hedgemon4, line 63: Add end of file line

Hedgemon4 (changes requested):

Hedgemon4 (changes requested):
See requested changes

Hedgemon4 (approved):

PR #382, opened 2022-08-15
methods-method_overloading-hard-q4

ketphan02:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/methods/method_overloading/hard/q4/RectangleOverload.java
```diff
@@ -0,0 +1,53 @@
+package methods.method_overloading.hard.q4;
+
+/*
+Write a Java Program that prompts the user to enter two integers and two doubles. Then write four overloaded methods
+called rectangleArea. The first takes both integer values for each side of the rectangle. The second takes the first
```
Hedgemon4, line 5: We should probably add something about the goal of the question being to calculate the area of the rectangles based on the inputs (I feel like it is inferred, but not specifically stated.

src/methods/method_overloading/hard/q4/RectangleOverload.java
```diff
@@ -0,0 +1,53 @@
...
+integer and the first double for each side of the rectangle. The third takes the second integer and uses it for all
+sides of the rectangle. The fourth takes the second double and uses it for all sides of the rectangle.
+
+Sample Output:
+  Enter two integers:
+  2 4
+  Enter two doubles:
+  5.3 3.3
+  The area of the rectangle is: 8
```
Hedgemon4, line 14: We should also maybe change these to be like area of the first rectangle, second, etc. so the students have a better idea of what order they should print them in.

src/methods/method_overloading/hard/q4/MainTest.java
```diff
@@ -0,0 +1,156 @@
...
+                Arguments.of(4, 76, 23.4246, 648372.1, 304, 93.6984, 5776, 4.2038638005841E11),
+                Arguments.of(32917, 4, 57.3, 1.321, 131668, 1886144.1, 16, 1.745041),
+                Arguments.of(-43, -23, 43, 763, 989, -1849.0, 529, 582169.0),
+                Arguments.of(4738, -32, -4372.432, 584.432391, -151616, -2.0716582816E7, 1024, 341561.21964997693)
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("twoIntegersInputProvider")
+    void twoIntegerCorrect(int a, int b, int area) throws Throwable {
```
Hedgemon4, line 98: These test names are bad

src/methods/method_overloading/hard/q4/MainTest.java
```diff
@@ -0,0 +1,156 @@
...
+        Object[][] arguments = {
+                {a, int.class},
+                {b, int.class}
+        };
+        MethodTest m = new MethodTest(RectangleOverload.class, "rectangleArea", arguments);
+        Object output = m.callMethod();
+        String errorMsg = "Your rectangleArea() method does not calculate the correct area for two integers.";
```
Hedgemon4, line 105: Should change the error messages to be "using two integers" instead of for (do this for all the messages)

src/methods/method_overloading/hard/q4/MainTest.java
```diff
@@ -0,0 +1,156 @@
...
+        String errorMsg = "Your rectangleArea() method does not calculate the correct area for one double.";
+        CustomAssertions._assertEquals(area, output, 0.0001, errorMsg);
+    }
+
+    @ParameterizedTest
+    @MethodSource("mainMethodInputProvider")
+    void printsCorrectMessage(int a, int b, double x, double y, int twoInts, double intDouble, int oneInt, double oneDouble) {
+        runWithInput(a + " " + b + " " + x + " " + y);
+
+        assertEquals(twoInts, Integer.parseInt(getItemByName("int_int")), "Your program does not print the correct message for the first area.");
```
Hedgemon4, line 151: Should be "... correct output for the area of the first rectangle". Apply this to all the error messages.

src/methods/method_overloading/hard/q4/RectangleOverload.java
```diff
@@ -0,0 +1,53 @@
+package methods.method_overloading.hard.q4;
+
+/*
+Write a Java Program that prompts the user to enter two integers and two doubles. Then write four overloaded methods
+called rectangleArea. The first takes both integer values for each side of the rectangle. The second takes the first
+integer and the first double for each side of the rectangle. The third takes the second integer and uses it for all
+sides of the rectangle. The fourth takes the second double and uses it for all sides of the rectangle.
```
Hedgemon4, line 7: Also, I think we should tell them to assume positive inputs.

src/methods/method_overloading/hard/q4/MainTest.java
```diff
@@ -0,0 +1,156 @@
...
+                Arguments.of(2, 2.2, 4.4),
+                Arguments.of(-1, 1.12, -1.12),
+                Arguments.of(-1, -1.0, 1.0),
+                Arguments.of(112, 2222.123, 248877.776));
+    }
+
+    static Stream<Arguments> oneDoubleInputProvider() {
+        return Stream.of(
+                Arguments.of(1.2, 1.44),
+                Arguments.of(-1.0, 1.0),
```
Hedgemon4, line 82: Get rid of the negative inputs

src/methods/method_overloading/hard/q4/RectangleOverload.java
```diff
@@ -0,0 +1,54 @@
+package methods.method_overloading.hard.q4;
+
+/*
+Write a Java Program that prompts the user to enter two integers and two doubles which will be used to calculate
```
Hedgemon4, line 4: We also need to specify the return type for the methods

Hedgemon4 (changes requested):

src/methods/method_overloading/hard/q4/RectangleOverload.java
```diff
@@ -0,0 +1,55 @@
...
+rectangle areas. This calculation will be performed by four overloaded methods called rectangleArea. The first takes
+both integer values for the sides of the rectangle and returns and integer. The second takes the first integer and the
```
Hedgemon4, line 6:
```suggestion
both integer values for the sides of the rectangle and returns an integer. The second takes the first integer and the
```

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #383, opened 2022-08-15
Randomness checking fix

Hedgemon4:
Fixed bug where random check was failing with random numbers and where RandomDouble was only generating one bucket


src/global/variables/clauses/RandomDouble.java
```diff
@@ -77,7 +77,9 @@ public boolean validateRandom(int matchGroupNum) {
     private static int getNumBins(double lower, double upper) {
         int upperCeil = (int) (upper + 1);
         int lowerFloor = (int) lower;
-        return RandomUtil.getNumBins(lowerFloor, upperCeil); // TODO: do properly
+        // If the number of bins would be less than call the method to ensure there are 10 bins
```
Hedgemon4, line 80:
```suggestion
        // Ensures that there will be at least 10 bins
```

PR #393, opened 2022-08-22
methods-method_overloading-medium-q5

ketphan02:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/methods/method_overloading/medium/q5/OverloadIt.java
```diff
@@ -0,0 +1,15 @@
+package methods.method_overloading.medium.q5;
+
+/*
+ Overload a Java Method so that it can take two integer variables or one integer variable and one double variable and
```
Hedgemon4, line 4: Fix this question up and let the student know the return type

src/methods/method_overloading/medium/q5/MainTest.java
```diff
@@ -0,0 +1,60 @@
...
+    @ParameterizedTest
+    @MethodSource("inputDoubleAndIntegerProvider")
+    void correctTwoDoubleProduct(double a, int b, double product) throws Throwable {
+        Object[][] arguments = {
+                {a, double.class},
+                {b, int.class}
+        };
+        MethodTest m = new MethodTest(OverloadIt.class, "multiply", arguments);
+        Object output = m.callMethod();
+        String errMsg = "Your multiply method does not return the correct product of 2 doubles.";
```
Hedgemon4, line 57: This is not two doubles

src/methods/method_overloading/medium/q5/MainTest.java
```diff
@@ -0,0 +1,60 @@
...
+    @ParameterizedTest
+    @MethodSource("inputTwoIntegersProvider")
+    void correctTwoIntegerProduct(int a, int b, int product) throws Throwable {
+        Object[][] arguments = {
+                {a, int.class},
+                {b, int.class}
+        };
+        MethodTest m = new MethodTest(OverloadIt.class, "multiply", arguments);
+        Object output = m.callMethod();
+        String errMsg = "Your multiply method does not return the correct product of 2 integers.";
```
Hedgemon4, line 44: I think correctly calculate or something similar would be better here

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #399, opened 2022-08-22
methods-method_overloading-hard-q6

ketphan02:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/methods/method_overloading/hard/q6/AttendingStudent.java
```diff
@@ -0,0 +1,57 @@
+package methods.method_overloading.hard.q6;
+
+/*
+Write a Java Program with an overloaded method called studentCheck to check whether a student has a valid username and
+student number. One method takes the first name, the last name, and the username and returns a boolean. The other name
```
Hedgemon4, line 5: Should be method, not name at the end of this sentence

src/methods/method_overloading/hard/q6/AttendingStudent.java
```diff
@@ -0,0 +1,57 @@
...
+takes the student number and returns a boolean. To determine this, ask the user for their name, username, and student
+number. The following conditions should be met to determine their validity:
+Their username must be their first name and last name back-to-back with a dot in the middle for example: john.doe
```
Hedgemon4, line 8: Also, I feel like these requirements are laid out funny and are unclear

src/methods/method_overloading/hard/q6/AttendingStudent.java
```diff
@@ -0,0 +1,57 @@
...
+    int len = String.valueOf(studentID).length();
+    boolean check;
+    if (len < 8)
+      check = false;
+    else
+      check = true;
+    return check;
+  }
+
+  public static boolean studentCheck(String fname, String lname, String username) {
```
Hedgemon4, line 48: Also, this is going to fail if capitalization is wrong. We should specify that if it is what we want

src/methods/method_overloading/hard/q6/MainTest.java
```diff
@@ -0,0 +1,152 @@
...
+                new StringLiteral(username),
+                new StringLiteral(String.valueOf(validName)),
+                new StringLiteral(String.valueOf(validNum))
+        });
+        assertEquals(Integer.parseInt(num), Integer.parseInt(getItemByName("studentNum")), "Your program does not print the given student number.");
+    }
+
+    @ParameterizedTest
+    @MethodSource("validStudentNumberProvider")
+    public void correctValidStudentCheckNumberMethod(int studentID) throws Throwable {
```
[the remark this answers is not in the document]
Hedgemon4: Also, I feel like the name could be better so they have a better idea of which of the overloaded methods we are checking

Hedgemon4 (changes requested):

src/methods/method_overloading/hard/q6/AttendingStudent.java
```diff
@@ -0,0 +1,56 @@
+package methods.method_overloading.hard.q6;
+
+/*
+Write a Java Program that asks a user for their name, username and student number and has an overloaded method called
+studentCheck that checks if the username or student number are valid. One method takes the first name, the last name,
```
Hedgemon4, line 5:
```suggestion
studentCheck that checks if the username and student number are valid. One method takes the first name, the last name,
```

src/methods/method_overloading/hard/q6/AttendingStudent.java
```diff
@@ -0,0 +1,56 @@
...
+and the username to check if the username follows the format of first_name.last_name, and returns a boolean. The other
+method takes the student number and checks it is 8-digits long, then returns a boolean. The username should be case
+sensitive.
```
Hedgemon4, line 8: You should also add that these should be printed in the main method

src/methods/method_overloading/hard/q6/MainTest.java
```diff
@@ -0,0 +1,131 @@
...
+        };
+    }
+
+    public void runMain() {
+        AttendingStudent.main(new String[0]);
+    }
+
+    static Stream<Arguments> studentNumberProvider() {
+        return Stream.of(
+                Arguments.of(12345678, true),
```
Hedgemon4, line 55: In the case they check the length using numbers instead of a string, you should add a case for 100000000 (which would be the first number above eight digits)

src/methods/method_overloading/hard/q6/MainTest.java
```diff
@@ -0,0 +1,131 @@
...
+    @ParameterizedTest
+    @MethodSource("mainMethodInputProvider")
+    public void correctMainMethodOutput(String name, String username, String num, boolean validName, boolean validNum) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not correctly identify valid usernames and student numbers.";
+        runWithInput(name + "\n" + username + "\n" + num, new Clause[]{
+                new StringLiteral(username),
+                new StringLiteral(String.valueOf(validName)),
+                new StringLiteral(String.valueOf(validNum))
+        });
+        assertEquals(Integer.parseInt(num), Integer.parseInt(getItemByName("studentNum")), "Your program does not print the given student number.");
```
Hedgemon4, line 103:
```suggestion
        assertEquals(Integer.parseInt(num), Integer.parseInt(getItemByName("studentNum")), "Your program does not print the input student number.");
```

Hedgemon4 (changes requested):

src/methods/method_overloading/hard/q6/MainTest.java
SamamthaH: aha whoops, looks like when I was rewriting the question I messed up the question a bit, it's supposed to say at least 8 digits not equal to 8 digits

src/methods/method_overloading/hard/q6/AttendingStudent.java
Hedgemon4: Fair

src/methods/method_overloading/hard/q6/MainTest.java
Hedgemon4: Ah all good then

Hedgemon4 (approved):

PR #401, opened 2022-08-22
methods-method_overloading-hard-q9

ketphan02:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/methods/method_overloading/hard/q9/Overload9.java
```diff
@@ -0,0 +1,34 @@
+package methods.method_overloading.hard.q9;
+
+/**
+ * [J] Write a Java Program that passes in a user's first name and age and
```
Hedgemon4, line 4: This question should be rewritten because we did not tell them the method name. Also, we might just be able to get rid of the main method in this one (but that is up to you). If you chose not to, we need to try and avoid the hardcoding though.

src/methods/method_overloading/hard/q9/MainTest.java
```diff
@@ -0,0 +1,65 @@
...
+        Arguments.of(24, "Matt", "Matt.24"),
+        Arguments.of(0, " ", " .0"),
+        Arguments.of(-1, "K", "K.-1"),
+        Arguments.of(1001, "", ".1001"),
+        Arguments.of(92, "Helen", "Helen.92"));
+  }
+
+  @ParameterizedTest
+  @MethodSource("inputNameAgeProvider")
+  void userNameWithNameAgeCorrect(String name, int age, String username) throws Throwable {
```
Hedgemon4, line 50: I feel like the test name is not clear

src/methods/method_overloading/hard/q9/MainTest.java
```diff
@@ -0,0 +1,65 @@
...
+    String errMsg = "Your method \"userName()\" does not give the correct username.";
+    Object output = MethodUtil.invokeIfMethodExists(Overload9.class, "userName", new Object[] { name, age },
+        String.class, int.class);
+    CustomAssertions._assertEquals(output, username, errMsg);
+  }
+
+  @ParameterizedTest
+  @MethodSource("inputAgeNameProvider")
+  void userNameWithAgeNameCorrect(int age, String name, String username) throws Throwable {
```
Hedgemon4, line 59: Nor is this

Hedgemon4 (changes requested):

src/methods/method_overloading/hard/q9/Overload9.java
Hedgemon4: We can also do the thing Bowen suggested though and give them the main method and just have them write the overloaded methods

Hedgemon4 (changes requested):

src/methods/method_overloading/hard/q9/Overload9.java
SamamthaH: sadly I had already finished converting this one to input test before the meeting happened

Hedgemon4: All good

src/methods/method_overloading/hard/q9/GenerateUsername.java
```diff
@@ -0,0 +1,40 @@
+package methods.method_overloading.hard.q9;
+
+/*
+Write a Java Program that receives two first names and ages from the user and passes that information to an
+overloaded method called userName. The method should take the String and int in either order, (name, age) or (age, name),
+and return their username as first_name.age. For example john.34.
```
Hedgemon4, line 6: I think you should elaborate on the example (if john and 34 are input, the user name is john.34 )

Hedgemon4 (commented):
See discord message

Hedgemon4 (approved):

PR #406, opened 2022-08-24
methods-modular_programs-medium-q9

ketphan02:
## PR Checklist
- [x]  EOF newlines
---
## Website


Hedgemon4 (commented):
See comments in other requests

src/methods/modular_programs/medium/q9/MaxItOut.java
```diff
@@ -0,0 +1,46 @@
+package methods.modular_programs.medium.q9;
+
+/*
+Write a program that asks for two integers and passes them to a method called maxOut which check is one of the numbers
```
Hedgemon4, line 4:
```suggestion
Write a program that asks for two integers and passes them to a method called maxOut which check if one of the numbers
```

src/methods/modular_programs/medium/q9/MaxItOut.java
```diff
@@ -0,0 +1,46 @@
...
+        int n2 = input.nextInt();
+        System.out.println(maxOut(n1, n2));
+    }
+
+    public static String maxOut(int n1, int n2) {
+        String res;
+        if (n1 % n2 == 0) {
+            if (n1 < n2)
+                res = n2 + " is the max";
+            else
```
Hedgemon4, line 33: I don't think you are ever going to need the else, because if a number is divisible by another number, the divisor must be less than the dividend. So if n1 % n2 == 0, then n1 is the max

Hedgemon4 (changes requested):

src/methods/modular_programs/medium/q9/MaxItOut.java
Hedgemon4: So you should be able to simplify the solution code

Hedgemon4 (changes requested):

src/methods/modular_programs/medium/q9/MaxItOut.java
SamamthaH: unfortunately, when one of the cases is 0 we do need the if statement

Hedgemon4 (approved):

PR #410, opened 2022-09-07
methods-modular_programs-hard-q9

ketphan02:
## PR Checklist
- [x]  EOF newlines
---
## Website


Hedgemon4 (changes requested):
See comments in other reqeusts

src/methods/modular_programs/hard/q9/TryAngleButReflected.java
```diff
@@ -0,0 +1,44 @@
...
+1 2 3 4 5 … n
+…
+1 2 3 4 5
+1 2 3 4
+1 2 3
+1 2
+1
+
+Sample output:
+Enter number of lines:
```
Hedgemon4, line 17: You should tab this over, I think it looks confusing otherwise

src/methods/modular_programs/hard/q9/TryAngleButReflected.java
```diff
@@ -0,0 +1,44 @@
...
+    public static void main(String[] args) {
+        Scanner input = new Scanner(System.in);
+        System.out.println("Enter number of lines: ");
+        int n = input.nextInt();
+        printPattern(n);
+    }
+
+    public static void printPattern(int n) {
+        if (n <= 0) {
+            System.out.println("Number must be positive");
```
Hedgemon4, line 36: Add a period or exclamation mark

src/methods/modular_programs/hard/q9/MainTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+    public void runMain() {
+        TryAngleButReflected.main(new String[0]);
+    }
+
+    static Stream<Integer> inputInvalidProvider() {
+        return Stream.of(0, -1, -2, -10, -100);
+    }
+
+    @RepeatedTest(50)
```
Hedgemon4, line 43: We might not want to have fifty tests each, that is a bit too much

Hedgemon4: Maybe code up like 10 test cases instead

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #422, opened 2022-09-17
arrays-creating_arrays-hard-q1

SamamthaH:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/arrays/creating_arrays/hard/q1/RandomArray.java
```diff
@@ -0,0 +1,42 @@
+package arrays.creating_arrays.hard.q1;
+
+/*
+Write a Java program that asks the user for the number of elements it will store and uses a method called
+createRandomArray() to create a new integer array and initializes the array with random numbers from 0 to the number
+that was entered. The main method should then print out the numbers in the array.
+Sample output:
```
Hedgemon4, line 7: Add space between question text and sample output

src/arrays/creating_arrays/hard/q1/RandomArray.java
```diff
@@ -0,0 +1,42 @@
...
+Enter the size of the array:
+5
```
Hedgemon4, line 9: Tab this over

src/arrays/creating_arrays/hard/q1/RandomArray.java
```diff
@@ -0,0 +1,42 @@
+package arrays.creating_arrays.hard.q1;
+
+/*
+Write a Java program that asks the user for the number of elements it will store and uses a method called
```
Hedgemon4, line 4: Change question text to:
Write a Java program that asks the user to enter a length for an array and then calls a method named createRandomArray() to generate this array. The method should accept an integer for the length and create a new array of that size. Then, it should initialize the array with random numbers from 0 (inclusive) to the passed length (exclusive). Return this array and print the output in the main method.

Hedgemon4 (changes requested):

SamamthaH:
I think the reason this is framework is because without the try catches we still don't have a good way of checking if the entire array is random ints

src/arrays/creating_arrays/hard/q1/MainTest.java
```diff
@@ -0,0 +1,132 @@
...
+                new PlaceHolder()
+        };
+    }
+
+    public void runMain() {
+        RandomArray.main(new String[0]);
+    }
+
+    public static Stream<Integer> createRandomArrayInputProvider() {
+        return Stream.of(1000, 1500, 2000, 3000);
```
Hedgemon4, line 45: We should probably try some other values that are not so large. With larger values, it is more likely they can pass the random check without actually doing it properly.

Hedgemon4 (changes requested):

src/arrays/creating_arrays/hard/q1/MainTest.java
```diff
@@ -0,0 +1,132 @@
...
+    @MethodSource("createRandomArrayInputProvider")
+    public void correctCreateRandomArrayMethod(int in) throws Throwable {
+        Object[][] arguments = {
+                {in, int.class}
+        };
+        MethodTest m = new MethodTest(RandomArray.class, "createRandomArray", arguments);
+        Object output = m.callMethod();
+        try {
+            int[] arr = (int[]) output;
+            ArrayList<Integer> response = Arrays.stream(arr).boxed().collect(Collectors.toCollection(ArrayList::new));
```
Hedgemon4, line 62: We should also verify that the array returned is the correct length

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #423, opened 2022-09-19
arrays-creating_arrays-hard-q2

SamamthaH:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/arrays/creating_arrays/hard/q2/RevdArray.java
```diff
@@ -0,0 +1,49 @@
+package arrays.creating_arrays.hard.q2;
+
+/*
+Write a program that will ask the user for 5 numbers and initialize an integer array with the reverse of each number.
+For example, if the user enters 432 as the 5th number, then the 5th array element will be 234. Assume only positive
+numbers will be input.
+To simplify your code, make sure you create a method called reverse() that takes a single integer as input and returns
+an integer in a reversed format as output.
```
Hedgemon4, line 8: Add space between question text and sample output

src/arrays/creating_arrays/hard/q2/RevdArray.java
```diff
@@ -0,0 +1,49 @@
...
+Sample output:
+Enter a number to add to the array:
```
Hedgemon4, line 10: Tab sample output over

src/arrays/creating_arrays/hard/q2/MainTest.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MainTest extends BaseTest {
+    // Java
+    public Clause[] testSentence() {
+        TestOption.isInputTest = true;
+        TestOption.defaultInput = "39 67 123 5 21";
+        return new Clause[]{
```
Hedgemon4, line 25: We may want to think about using a `ClauseBuilder()` here so that we can more easily vary the input on the website.

Hedgemon4: It would also just make the code cleaner

src/arrays/creating_arrays/hard/q2/MainTest.java
```diff
@@ -0,0 +1,95 @@
...
+                Arguments.of("0 55 64 13 7356", 0, 55, 46, 31, 6537),
+                Arguments.of("319 404 257 123 666", 913, 404, 752, 321, 666),
+                Arguments.of("10000 43 7 8273 4372", 1, 34, 7, 3728, 2734)
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("reverseInputProvider")
+    void correctReverseMethod(int in, int ans) throws Throwable {
+        Object output = MethodUtil.invokeIfMethodExists(RevdArray.class, "reverse", new Object[]{in}, int.class);
```
Hedgemon4, line 81: Switch this to methodTest

src/arrays/creating_arrays/hard/q2/MainTest.java
```diff
@@ -0,0 +1,95 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MainTest extends BaseTest {
+    // Java
```
Hedgemon4, line 21: Add a space between this tag and the main method

src/arrays/creating_arrays/hard/q2/MainTest.java
```diff
@@ -0,0 +1,95 @@
...
+    void correctReverseMethod(int in, int ans) throws Throwable {
+        Object output = MethodUtil.invokeIfMethodExists(RevdArray.class, "reverse", new Object[]{in}, int.class);
+        CustomAssertions._assertEquals(ans, output, "Your reverse method does not correctly reverse the given number.");
+    }
+
+    @ParameterizedTest
+    @MethodSource("mainMethodInputProvider")
+    void printsOutputCorrectly(String in, int one, int two, int three, int four, int five) {
+        runWithInput(in);
+        assertEquals(String.valueOf(one), getItemByName("one"), "Your program does not correctly reverse the first number");
```
Hedgemon4, line 89: You should do this the other way around and assert them as ints, not strings

Hedgemon4 (changes requested):

src/arrays/creating_arrays/hard/q2/RevdArray.java
```diff
@@ -0,0 +1,49 @@
+package arrays.creating_arrays.hard.q2;
+
+/*
+Write a program that will ask the user for 5 numbers and initialize an integer array with the reverse of each number.
```
Hedgemon4, line 4: Change question text to:
`Write a program asking the user for five numbers and initialize an integer array with the reverse of each number. For example, if the user enters 432 as the 5th number, then the 5th array element will be 234. Assume only positive numbers will be input. To simplify your code, create a method called reverse() that takes a single integer as input and returns an integer in a reversed format as output.`

Hedgemon4 (changes requested):

src/arrays/creating_arrays/hard/q2/RevdArray.java
Hedgemon4: Sample output is not tabbed over

Hedgemon4 (changes requested):
See comments

src/arrays/creating_arrays/hard/q2/RevdArray.java
```diff
@@ -0,0 +1,49 @@
+package arrays.creating_arrays.hard.q2;
+
+/*
+Write a program asking the user for five numbers and initialize an integer array with the reverse of each number. For
+example, if the user enters 432 as the 5th number, then the 5th array element will be 234. Assume only positive numbers
+will be input. To simplify your code, create a method called reverse() that takes a single integer as input and returns
+an integer in a reversed format as output.
```
Hedgemon4, line 7:
```suggestion
the reverse of that integer. Assume the user will only input positive integers. 
```

src/arrays/creating_arrays/hard/q2/MainTest.java
```diff
@@ -0,0 +1,105 @@
...
+        CustomAssertions._assertEquals(ans, output, "Your reverse method does not correctly reverse the given number.");
+    }
+
+    @ParameterizedTest
+    @MethodSource("mainMethodInputProvider")
+    void printsOutputCorrectly(int[] in) {
+        runWithInput(ArrayUtil.arrayToInput(in));
+
+        for(int x = 0; x < in.length; x++){
+            assertEquals(solution(in[x]), Integer.parseInt(getItemByName("num" + x)), "Your program does not correctly reverse array element" + x);
```
Hedgemon4, line 75: I don't think you need to specify the array element, I think you can make the message a bit more generic. 
```suggestion
            assertEquals(solution(in[x]), Integer.parseInt(getItemByName("num" + x)), "Your program does not correctly reverse each element in the array." + x);
```

src/arrays/creating_arrays/hard/q2/MainTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @ParameterizedTest
+    @MethodSource("reverseInputProvider")
+    void correctReverseMethod(int in, int ans) throws Throwable {
+        Object[][] arguments = {
+                {in, int.class}
+        };
+        MethodTest m = new MethodTest(RevdArray.class, "reverse", arguments);
+        Object output = m.callMethod();
+        CustomAssertions._assertEquals(ans, output, "Your reverse method does not correctly reverse the given number.");
```
Hedgemon4, line 66:
```suggestion
        CustomAssertions._assertEquals(ans, output, "Your reverse method does not correctly reverse the input integer.");
```

Hedgemon4 (changes requested):

src/arrays/creating_arrays/hard/q2/MainTest.java
```diff
@@ -0,0 +1,108 @@
...
+import global.variables.wrappers.Optional;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MainTest extends BaseTest {
```
Hedgemon4, line 21: We might want to consider changing this one to a Parson's question, as I don't think we can actually guarantee that they are using an array.

Hedgemon4 (changes requested):

src/arrays/creating_arrays/hard/q2/RevdArray.java
```diff
@@ -0,0 +1,48 @@
+package arrays.creating_arrays.hard.q2;
+
+/*
+Write a program asking the user for ten numbers and stores them in an integer array. Then use a method called reverse
```
Hedgemon4, line 4: I don't know if we can have this question where the method includes dealing with an array here because it is before the arrays with methods section.

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #426, opened 2022-09-19
Add RandomBoolean Clause

Hedgemon4:


src/global/variables/clauses/RandomBoolean.java
```diff
@@ -0,0 +1,57 @@
...
+        return Boolean.parseBoolean(matchGroupString);
+    }
+
+    public ArrayList<Boolean> getValuesForMatchGroup(int matchGroup) {
+        // Maps Integer ArrayList back to the input boolean values to be displayed in the event of an error
+        return valueStore.get(matchGroup).stream().map(e -> (e == 1)).collect(Collectors.toCollection(ArrayList::new));
+    }
+
+    public String getRegex() {
+        return "((?:true)|(?:false))";
```
keyvankhademi, line 55: It's good if we can specify the choices. the default can be "true" or "false".

src/global/utils/RandomUtil.java
```diff
@@ -106,9 +107,19 @@ public static boolean frequenciesAreRandom(int[] frequencies, int numBins) {
         return (count * 1.0) / numBins >= ACCEPTANCE_RATE;
     }
 
+    public static boolean frequenciesAreRandom(int[] frequencies, double percentageTrue) {
```
keyvankhademi, line 110: Why can't we use the current function with numBins = 2?

keyvankhademi (changes requested):

src/global/variables/clauses/RandomBoolean.java
Hedgemon4: Why do we want to be able to specify the choices? Since it tests if half is true, and half is false, there would not be a scenario where we would just specify one of the two.

src/global/utils/RandomUtil.java
Hedgemon4: That is because we only need to really test the one bin, and the error acceptance is much lower for booleans than it is for everything else.

src/global/variables/clauses/RandomBoolean.java
keyvankhademi: no i mean what if the output was 0 and 1 instead of "true" or "false"

PR #427, opened 2022-09-23
Created MethodTest so the Clause system can be used when testing methods

Hedgemon4:


src/test/MethodCodeTest.java
```diff
@@ -75,9 +75,9 @@ void checkGetMethodOutput() throws Throwable {
 //        MethodUtil.invokeIfMethodExists(MethodCode.class, "ThisMethod");
 //    }
 
-    @Test
```
keyvankhademi, line 78: Why is this commented out?

src/test/MethodClauseTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @ParameterizedTest
+    @MethodSource("areaCalcInputProvider")
+    void correctAreaMethod(double x, double y, double area) throws Throwable {
+        MethodTest m = new MethodTest(MethodClause.class, "areaCalc", new Object[]{x, y},
+                new Class[]{double.class, double.class}, new Clause[]{
+                new StringLiteral("The area is "),
+                new DoubleLiteral("area")
+        });
+        //TestOption.incorrectMethodStructureErrorMessage = "Your areaCalc method does not correctly print out the area.";
```
keyvankhademi, line 81: This is an extra comment I believe.

src/global/utils/MethodUtil.java
```diff
@@ -11,11 +10,15 @@
...
+/**
+ * @deprecated This class should no longer be used for testing methods. Instead, the new MethodTest class should be used.
+ */
+@Deprecated
 public class MethodUtil {
     private static ByteArrayOutputStream methodOutput;
 
     public static Object invokeIfMethodExists(Class<?> methodClass, String methodName, Object[] arguments, Class<?>... methodArgumentTypes) throws Throwable {
-        setUpMethodOutput();
+        methodSetUp();
```
keyvankhademi, line 21: We don't need this name change anymore I think.

src/test/MethodClauseTest.java
```diff
@@ -0,0 +1,103 @@
...
+                new NewLine(),
+                new StringLiteral("Hello world"),
+                new NewLine(),
+                new DoubleLiteral(30, 35, "double")
+        };
+        MethodTest m = new MethodTest(MethodClause.class, "testMethod", c, "Your double value is outside the range.");
+        m.setMethodNotFoundErrorMessage("Test ");
+        Object output = m.invokeIfMethodExists();
+        Logger.logMessage(m.getMethodOutput());
+        assertEquals(35, Integer.parseInt(m.getMethodItemByName("integer")), "ints were different");
```
keyvankhademi, line 99: Are these the messages we show to the user? Maybe they should be more detailed.

src/test/MethodClauseTest.java
```diff
@@ -0,0 +1,103 @@
...
+        m.invokeIfMethodExists();
+    }
+
+    @ParameterizedTest
+    @MethodSource("addInputProvider")
+    void correctAddMethod(int x, int y, int z) throws Throwable {
+        MethodTest m = new MethodTest(MethodClause.class, "add", new Object[]{x, y},
+                new Class[]{int.class, int.class});
+        Object output = m.invokeIfMethodExists();
+        CustomAssertions._assertEquals(z, output, "Your add method does not correctly add two numbers.");
```
keyvankhademi, line 70: Does the user see which method we are referring to? If not maybe we should be more specific in the message.

keyvankhademi: I know this is not a real question but when we will ask people to use this as an example of writing method tests. It's good to have it show exactly what we expect in the junits.

keyvankhademi (changes requested):

src/global/MethodTest.java
```diff
@@ -0,0 +1,241 @@
...
+        this.methodClass = methodClass;
+        this.methodName = methodName;
+        this.arguments = null;
+        this.methodArgumentTypes = null;
+        this.methodTestSentence = methodTestSentence;
+        this.incorrectMethodStructureErrorMessage = incorrectMethodStructureErrorMessage;
+        setMethodOutput();
+    }
+
+    public Object invokeIfMethodExists() throws Throwable {
```
Opeyem1a, line 114: I feel like we could just call this `callMethod()` or something so it makes more sense from a test write perspective. The term "invoke" only makes sense because I know that's what's under the hood but ideally test writers aren't looking at that.

You could implement this by leaving this method as is and just adding another method:
```java
public Object callMethod() throws Throwable {
    return invokeIfMethodExists();
}
```

Opeyem1a (changes requested):

src/global/MethodTest.java
Hedgemon4: I can also just change to rename the Method now because we are using a new file now to do the testing entirely.

src/global/MethodTest.java
```diff
@@ -0,0 +1,241 @@
...
+        this.methodClass = methodClass;
+        this.methodName = methodName;
+        this.arguments = null;
+        this.methodArgumentTypes = null;
+        this.methodTestSentence = null;
+        this.incorrectMethodStructureErrorMessage = "Your " + methodName + " method's output did not follow the correct structure/syntax";
+        setMethodOutput();
+    }
+
+    public MethodTest(Class<?> methodClass, String methodName, Object[] arguments, Class<?>[] methodArgumentTypes, Clause[] methodTestSentence) {
```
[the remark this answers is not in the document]
Hedgemon4: @keyvankhademi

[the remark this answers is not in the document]
keyvankhademi: Hmm! This definitely looks better when calling the method. On the other hand it might make the implementation more dirty (maybe not).

@Hedgemon4 Let's go for it.

src/test/MethodClauseTest.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @ParameterizedTest
+    @MethodSource("areaCalcInputProvider")
+    void correctAreaMethod(double x, double y, double area) throws Throwable {
+        MethodTest m = new MethodTest(MethodClause.class, "areaCalc", new Object[]{x, y},
+                new Class[]{double.class, double.class}, new Clause[]{
```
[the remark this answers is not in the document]
keyvankhademi: +

src/test/MethodClauseTest.java
Hedgemon4: The user should know what method we are referring to because general if the method is called `add(int x, int y)` we will say that `Your add method does not correctly add two numbers.` .

src/test/MethodClauseTest.java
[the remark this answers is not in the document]
Hedgemon4: Ok sounds good, we can have that be the standard instead of how it is done here. This was mostly just a quick test that I wrote so that I knew everything was working, but I can flush it out so we have a better example.

src/test/MethodClauseTest.java
```diff
@@ -0,0 +1,103 @@
...
+                new NewLine(),
+                new DoubleLiteral(30, 35, "double")
+        };
+        MethodTest m = new MethodTest(MethodClause.class, "testMethod", c, "Your double value is outside the range.");
+        m.setMethodNotFoundErrorMessage("Test ");
+        Object output = m.invokeIfMethodExists();
+        Logger.logMessage(m.getMethodOutput());
+        assertEquals(35, Integer.parseInt(m.getMethodItemByName("integer")), "ints were different");
+        assertEquals(34.124, Double.parseDouble(m.getMethodItemByName("double")), "double diff");
+        CustomAssertions._assertEquals(69, output, "method output incorrect");
```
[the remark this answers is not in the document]
Hedgemon4: Sounds good, I will add this

src/test/MethodClause.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public static void print(String s) {
+        System.out.println(s);
+    }
+
+    public static int add(int x, int y) {
+        return x + y;
+    }
+
+    public static double areaCalc(double x, double y) {
```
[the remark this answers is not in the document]
Hedgemon4: Right, I can add those comments so the reader knows what these examples are for and how the return and printed output should be independent of each other.

src/test/MethodCodeTest.java
Hedgemon4: Probably because I was just testing something and I just wanted to enable you and Opey to be able to review the source MethodTest code

src/test/MethodClauseTest.java
Hedgemon4: Yeah it is

src/global/MethodTest.java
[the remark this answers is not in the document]
Hedgemon4: Sounds good, I will work on designing this.

src/test/MethodCodeTest.java
keyvankhademi: Should we change it back now?

Hedgemon4: It was because I was just using them to make sure errors were being thrown when they should

keyvankhademi (approved):

PR #429, opened 2022-09-23
arrays-creating_arrays-hard-q5

SamamthaH:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/arrays/creating_arrays/hard/q5/AlphabetRays.java
```diff
@@ -0,0 +1,22 @@
...
+
+/*
+Write a Java program called AlphabetRays to initialize an integer array with all of the lowercase letters of the English
+alphabet and print each character on a separate line. Recall that the character a has the integer equivalent value of 97.
+Sample output:
+a
+b
+…
+y
+z
```
Hedgemon4, line 11: Fix the formatting of the sample output

src/arrays/creating_arrays/hard/q5/AlphabetRays.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+public class AlphabetRays {
+    public static void main(String[] args) {
```
Hedgemon4, line 15: Actually, we should maybe talk about questions like this because theoretically they could print the letters and not put them into an array.

Hedgemon4: We also do not have a good way of testing if they actually create the array.

Hedgemon4 (changes requested):

Hedgemon4 (approved):
This was a good way of redoing this question.

PR #430, opened 2022-09-23
arrays-creating_arrays-hard-q4

SamamthaH:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/arrays/creating_arrays/hard/q4/MainTest.java
```diff
@@ -0,0 +1,72 @@
...
+        };
+    }
+
+    public void runMain(){
+        RandomPickingNames.main(new String[0]);
+    }
+
+    static Stream<Arguments> inputProvider(){
+        return Stream.of(
+                Arguments.of("Obi-wan \n Darth Revan \n Ashoka \n Luke \n Jaba the Hutt", 0),
```
Hedgemon4, line 53:
```suggestion
                Arguments.of("Obi\\-wan \n Darth Revan \n Ashoka \n Luke \n Jaba the Hutt", 0),
```

Hedgemon4 (changes requested):

src/arrays/creating_arrays/hard/q4/MainTest.java
```diff
@@ -0,0 +1,78 @@
+package arrays.creating_arrays.hard.q4;
```
Hedgemon4, line 1: I am fairly certain this one will also have a problem with if they use an array or not, so we should talk about it.

Hedgemon4: Though we might be able to make using an array very difficult by adding scanner input specifying the number of names to be entered.

Hedgemon4 (changes requested):

src/arrays/creating_arrays/hard/q4/RandomPickingNames.java
```diff
@@ -0,0 +1,32 @@
+package arrays.creating_arrays.hard.q4;
+
+/*
+Write a program that asks the user for 5 names of students, stores those values into an array, and asks the user for
```
Hedgemon4, line 4:
```suggestion
Write a program that asks the user for 50 names of students, stores those values in an array, and asks the user for
```

src/arrays/creating_arrays/hard/q4/MainTest.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void picksRandomNameCorrectly(String[] in, int c) throws InvalidClauseException {
+        runWithInput(arrayToInput(in) + c, new Clause[]{
+                new StringLiteral(in[c], "name")
+        });
+
+        assertEquals(c, Integer.parseInt(getItemByName("index")), "Your program does not correctly print the given index.");
```
Hedgemon4, line 80:
```suggestion
        assertEquals(c, Integer.parseInt(getItemByName("index")), "Your program does not print the input index.");
```

src/arrays/creating_arrays/hard/q4/MainTest.java
```diff
@@ -0,0 +1,83 @@
...
+        assertEquals(in[c], getItemByName("name"), "Your program does not print the correct name located at the given index.");
```
Hedgemon4, line 81:
```suggestion
        assertEquals(in[c], getItemByName("name"), "Your program does not print the element at the input index.");
```

Hedgemon4 (approved):

PR #434, opened 2022-09-26
arrays-arrays_with_methods-medium-q9

SamamthaH:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/arrays/arrays_with_methods/medium/q9/TwoDAndThatsAFact.java
```diff
@@ -0,0 +1,37 @@
+package arrays.arrays_with_methods.medium.q9;
+
+/*
+Make a program that will calculate the factorial of each element in a 2D array and store the factorial of each element
```
Hedgemon4, line 4: I think we should tweak this question text to mention we give an array in the main method and print the output of that generated 2D array.

Hedgemon4 (changes requested):

src/arrays/arrays_with_methods/medium/q9/TwoDAndThatsAFact.java
```diff
@@ -0,0 +1,36 @@
+package arrays.arrays_with_methods.medium.q9;
+
+/*
+Make a program that defines a 2D array in the main method. Then calculate the factorial of each element using a method
+called fact that returns an integer, and store it in a separate 2D array. Print the resulting array.
```
Hedgemon4, line 5:
```suggestion
called fact that returns an integer, and stores it in a separate 2D array. Print the resulting array.
```

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #437, opened 2022-10-02
arrays-arrays_with_methods-hard-q10

SamamthaH:
## PR Checklist
- [x]  EOF newlines
---
## Website


Hedgemon4 (changes requested):
Update test cases to use arrayUtil

SamamthaH:
> Update test cases to use arrayUtil

Is that really necessary? The point of arrayUtil is that it can make very large test cases, but I already have a test case in this that is 1000 size

src/arrays/arrays_with_methods/hard/q10/ArrayofDigits.java
```diff
@@ -0,0 +1,34 @@
+package arrays.arrays_with_methods.hard.q10;
+
+/*
+Write a Java program that asks the user to enter two integers, where the first number n represents the numeric value to
+be stored and the second number k represents the repetitions of the first number. Then, call a method that you will also
+need to define called digitInt, which returns an integer array of size k all initialized to the value of n. Lastly,
+in the main method, display the resulting array that was outputted from the method call.
```
Hedgemon4, line 7:
```suggestion
in the main method, print the array which is returned from the digitInt method.
```

src/arrays/arrays_with_methods/hard/q10/MainTest.java
```diff
@@ -0,0 +1,87 @@
...
+                {b, int.class}
+        };
+        MethodTest m = new MethodTest(ArrayofDigits.class, "digitInt", arguments);
+        Object output = m.callMethod();
+
+        int[] arr = new int[b];
+        for (int x = 0; x < b; x++)
+            arr[x] = a;
+
+        CustomAssertions._assertArrayEquals(arr, output, "Your digitInt method does not return an array of the specified integers to the specified length.");
```
Hedgemon4, line 72:
```suggestion
        CustomAssertions._assertArrayEquals(arr, output, "Your digitInt method does not return the correct array based on the input parameters.");
```

src/arrays/arrays_with_methods/hard/q10/MainTest.java
```diff
@@ -0,0 +1,87 @@
...
+    @ParameterizedTest
+    @MethodSource("mainMethodInputProvider")
+    void correctMainMethodOutput(String in, int a, int b) throws InvalidClauseException {
+        Clause[][] c = new Clause[1][b * 2];
+        int count = 0;
+        for (int x = 0; x < b; x++) {
+            c[0][count++] = new IntegerLiteral(a);
+            c[0][count++] = new NewLine();
+        }
+
```
Hedgemon4, line 84: We should add a custom error message here I think

Hedgemon4 (changes requested):

src/arrays/arrays_with_methods/hard/q10/MainTest.java
```diff
@@ -0,0 +1,88 @@
...
+        for (int x = 0; x < b; x++)
+            arr[x] = a;
+
+        CustomAssertions._assertArrayEquals(arr, output, "Your digitInt method does not return the correct based on the input parameters.");
+    }
+
+    @ParameterizedTest
+    @MethodSource("mainMethodInputProvider")
+    void correctMainMethodOutput(String in, int a, int b) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not correctly print the array of integers.";
```
Hedgemon4, line 78:
```suggestion
        TestOption.incorrectStructureErrorMessage = "Your program does not correctly print the generated array of integers.";
```

Hedgemon4 (changes requested):

PR #439, opened 2022-10-03
arrays-arrays_with_methods-hard-q12

SamamthaH:
## PR Checklist
- [x]  EOF newlines
---
## Website


Hedgemon4 (changes requested):
Update test cases to use arrayUtil

SamamthaH:
> Update test cases to use arrayUtil

Why? We don't pass arrays as inputs in this question.

src/arrays/arrays_with_methods/hard/q12/MainTest.java
```diff
@@ -0,0 +1,62 @@
...
+import global.variables.wrappers.Optional;
+import org.junit.jupiter.api.RepeatedTest;
+import org.junit.jupiter.api.RepetitionInfo;
+
+public class MainTest extends BaseTest {
+    // Java
+
+    int[] factorials = {1, 2, 6, 24, 120, 720, 5040, 40320, 362880, 3628800, 39916800, 479001600, 1932053504, 1278945280, 2004310016};
+
+    public Clause[] testSentence() {
```
Hedgemon4, line 22: I think I asked for arrayUtil in this case by accident. I meant that we should maybe have a larger test case if possible so we can make sure that they are calculating factorial and not just using a switch or something.

src/arrays/arrays_with_methods/hard/q12/ArrayOfFacts.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+import java.util.Scanner;
+
+public class ArrayOfFacts {
+    public static void main(String[] args) {
+        Scanner input = new Scanner(System.in);
+        System.out.println("Enter the size of the array: ");
+        int n = input.nextInt();
+        int[] arr = new int[n];
```
Hedgemon4, line 27: I don't think we can actually make sure that they use an array here because they can just get the factorial and print it out. We should probably change this question.

Hedgemon4 (changes requested):

src/arrays/arrays_with_methods/hard/q12/ArrayOfMultiples.java
```diff
@@ -0,0 +1,43 @@
+package arrays.arrays_with_methods.hard.q12;
+
+/*
+Write a Java program that asks the user to input an array of 10 integers then passes it to a method called
+multiplyElements. The method should return a new integer array containing the element stored in the original array
+multiplied by the preceding element, with the first element wrapping back around to be multiplied by the last. Ex.
+element 2 in the new array is element 2 * element 1 from the original array. The new array should be returned and
+printed in the main method. Note: your multiplyElements method should be general enough to work with any array size.
```
Hedgemon4, line 8: For this, we should specify that first element should get multiplied by the last element if that is your intention

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #446, opened 2022-10-10
arrays-programs_involving_data_sequences-easy-q6

SamamthaH:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/arrays/programs_involving_data_sequences/easy/q6/MainTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+public class MainTest extends BaseTest {
+    // Parsons
+
+    //vary this variable to vary number of bmis calculated. Max of 20 (inclusive) because of default input.
+    final static int size = 10;
+
+    public Clause[] testSentence() {
+        TestOption.isInputTest = true;
+        TestOption.defaultInput = "180 65 135 45 187.3 95 67 23 132 78.3 184.78 67.521 704 15.4 63.5 38 374 23.4 437 38 180 65 135 45 187.3 95 67 23 132 78.3 184.78 67.521 704 15.4 63.5 38 374 23.4 437 38";
```
Hedgemon4, line 28: We should be able to just generate this default input with arrayUtil and then we don't need this limit on the variation,

Hedgemon4: As it doesn't really matter what the default is

src/arrays/programs_involving_data_sequences/easy/q6/MainTest.java
```diff
@@ -0,0 +1,116 @@
...
+        StringBuilder s = new StringBuilder();
+        for (int x = 0; x < size * 2; x++) {
+            s.append(values[x]);
+            s.append(" ");
+        }
+
+        runWithInput(s.toString());
+
+        for (int x = 0; x < size; x++) {
+            assertEquals(bmis[x], Double.parseDouble(getItemByName(String.valueOf(x + 1))), 0.00000001, "Your program calculate BMI " + (x + 1) + " correctly.");
```
Hedgemon4, line 89: Generally, I don't think we need to tell them at what specific line their code messed up

Hedgemon4 (changes requested):

src/arrays/programs_involving_data_sequences/easy/q6/BMWhy.java
```diff
@@ -0,0 +1,42 @@
...
+    BMI 9: 101.65289256198345
+    BMI 10: 9.192760700948004E-4
+ */
+
+public class BMWhy {
+    public static double calculateBMI(double height, double weight) {
+        return weight / (height * height);
+    }
+
+    public static void main(String[] args) {
```
Hedgemon4, line 28: Similar thing here with this question. We should change the question so that we know they are actually using an array.

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #450, opened 2022-10-17
conditionals-programs_with_decision_points-hard-q11

KaiKhademi:
## PR Checklist
- [x]  EOF newlines
---
## Website

### DIFF
--- a/src/conditionals/programs_with_decision_points/hard/q11/Discount.java
+++ b/src/conditionals/programs_with_decision_points/hard/q11/Discount.java
@@ -1,0 +1,34 @@
+package conditionals.programs_with_decision_points.hard.q11;
+/*
+Write a Java question that helps a store cashier determine if a given purchase qualifies for the birthday discount.
+To promote customer loyalty, the store gives all its customers a 10% discount for every purchase within 45 days of their
+birthday. In this program, ask the user for the birthday (day and month only) as integers and calculates if today is 45
+days from that birthday or not. The program should display whether the purchase qualifies for a 10% discount or not.
+ Assume the current date is 27/09 and all months have 30 days in them. If the person qualifies for the discount, print
+"You qualify for a 10% discount!", and if not print "You do not qualify for any discounts.".
+Sample Output:
+	Enter your birthday:
+    12
+    Enter your birth month as a number:
+    11
+    You qualify for a 10% discount!
+
+ */
+import java.util.Scanner;
+public class Discount {
+    public static void main(String[] args) {
+        Scanner input = new Scanner(System.in);
+        System.out.println("Enter your birth day: ");
+        int birthDay = input.nextInt();
+        System.out.println("Enter your birth month as a number: ");
+        int birthMonth = input.nextInt();
+        int currentMonth = 9;
+        int currentDay = 27;
+        int monthDiff = currentMonth - birthMonth;
+        int dayDiff = currentDay - birthDay;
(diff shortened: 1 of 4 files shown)


src/conditionals/programs_with_decision_points/hard/q11/Discount.java
```diff
@@ -0,0 +1,34 @@
+package conditionals.programs_with_decision_points.hard.q11;
+/*
+Write a Java question that helps a store cashier determine if a given purchase qualifies for the birthday discount.
+To promote customer loyalty, the store gives all its customers a 10% discount for every purchase within 45 days of their
+birthday. In this program, ask the user for the birthday (day and month only) as integers and calculates if today is 45
+days from that birthday or not. The program should display whether the purchase qualifies for a 10% discount or not.
+ Assume the current date is 27/09 and all months have 30 days in them. If the person qualifies for the discount, print
+"You qualify for a 10% discount!", and if not print "You do not qualify for any discounts.".
```
Hedgemon4, line 8: Add a space between the question wording and the sample output

src/conditionals/programs_with_decision_points/hard/q11/Discount.java
```diff
@@ -0,0 +1,34 @@
...
+Sample Output:
+	Enter your birthday:
```
Hedgemon4, line 10: Fix the spacing here so this is aligned with the rest of the text

src/conditionals/programs_with_decision_points/hard/q11/Discount.java
```diff
@@ -0,0 +1,34 @@
+package conditionals.programs_with_decision_points.hard.q11;
+/*
+Write a Java question that helps a store cashier determine if a given purchase qualifies for the birthday discount.
```
Hedgemon4, line 3: Change question text to:
`Write a Java question that helps a store cashier determine if a purchase qualifies for the birthday discount. To promote customer loyalty, the store gives all its customers a 10% discount for all purchases within 45 days of their birthday. Ask the user for their birthday (day and month only) as integers and calculates if today is 45 days from that day. The program should display whether the purchase qualifies for a 10% discount or not. Assume the current date is 27/09 and all months have 30 days. If the person qualifies for the discount print out "You qualify for a 10% discount!", and if not, print "You do not qualify for any discounts.".`

src/conditionals/programs_with_decision_points/hard/q11/MainTest.java
```diff
@@ -0,0 +1,89 @@
...
+            for (int j = 1; j <= 30; j++) {
+                if (!calcAns(j, i))
+                    unqualifieds[t++] = new int[] {j, i};
+            }
+        }
+        return Stream.of(unqualifieds);
+    }
+    @ParameterizedTest
+    @MethodSource("InputProviderUnqualified")
+    void testWithInputUnqualified(int[] in) throws InvalidClauseException {
```
Hedgemon4, line 83: As I mentioned in the onboarding, all test method names are eventually shown to the student as part of the feedback process. When developing these, it is fine to start the method names out like this, but you should change them to something that makes more sense to the student. In this case, if their code fails, they will just receive a message like `What went wrong: testWithInputUnqualifed failed, "Your program did not display the correct answer" ` which is not very specific.

Hedgemon4 (changes requested):
Some of these comments are probably present in your other pull requests, so please apply anything I mentioned here to those other ones as well.

PR #454, opened 2022-10-20
arrays-arrays_with_methods-hard-q6

SamamthaH:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/arrays/arrays_with_methods/hard/q6/MainTest.java
```diff
@@ -0,0 +1,73 @@
...
+public class MainTest {
+    // Java
+
+    public static Stream<Integer> randomReplaceInput(){
+        return Stream.of(3, 4, 5, 6, 7, 50);
+    }
+
+    @ParameterizedTest
+    @MethodSource("randomReplaceInput")
+    public void correctRandomReplaceMethod(int size) throws Throwable {
```
Hedgemon4, line 24: You don't need this return type, the custom assertion does this for you.

src/arrays/arrays_with_methods/hard/q6/MainTest.java
```diff
@@ -0,0 +1,73 @@
...
+        int repetitions = 0;
+        //get first round of numbers
+        for(int x = 0; x < size; x++){
+            if( x % 2 == 0 )
+                storeRandoms[count++][repetitions] = hold[x];
+        }
+        repetitions++;
+
+        //now rest of numbers
+        for(int x = 0; x < 999; x ++){
```
Hedgemon4, line 51: Also, we should separate these tests out and use an input provide like we normally do

Hedgemon4 (changes requested):

src/arrays/arrays_with_methods/hard/q6/MainTest.java
SamamthaH: we have a getMethodReturnType

src/arrays/arrays_with_methods/hard/q6/MainTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+public class MainTest {
+    // Java
+
+    public static Stream<Integer> randomReplaceInput() {
+        return Stream.of(3, 4, 5, 6, 7, 50);
+    }
+
+    @Test
+    public void randomReplaceMethodKeepsOddIndexNumber() throws Throwable {
```
Hedgemon4, line 27: This should probably be a parameterized test, as we should check this multiple times.

Hedgemon4 (changes requested):

src/arrays/arrays_with_methods/hard/q6/MainTest.java
```diff
@@ -0,0 +1,97 @@
...
+        int repetitions = 0;
+
+        //runs the program 1000 times and stores all the even index numbers generated. Each are in their own array
+        //because each index has its own range the variables can fall in (I'm effectively using ascending array to control that)
+        for (int x = 0; x < 1000; x++) {
+            for (int y = 0; y < size; y++) {
+                arr[y] = y + 1;
+            }
+            int count = 0;
+            int[] output = (int[]) m.callMethod();
```
Hedgemon4, line 80: If they don't return an array, will this not crash the program?

Hedgemon4 (changes requested):

src/arrays/arrays_with_methods/hard/q6/MainTest.java
SamamthaH: nope, the above
if (m.getMethodReturnType() != int[].class) {
            Assertions.fail("Your createRandomArray method does not return an array of integers.");
        }
is to make sure that if they don't return an int[] that the test fails before it crashes. Just tested it by making the method return void and the tests just returned a lot of "Your createRandomArray method does not return an array of integers." with no crashes

Hedgemon4 (approved):

PR #459, opened 2022-10-25
arrays-programs_involving_data_sequences-medium-q7

SamamthaH:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/arrays/programs_involving_data_sequences/medium/q7/MainTest.java
```diff
@@ -0,0 +1,61 @@
...
+import global.variables.Clause;
+import global.variables.clauses.StringLiteral;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
+
+public class MainTest extends BaseTest {
+    //Parsons with distractors
```
Hedgemon4, line 16:
```suggestion
    // Parsons with distractors
```

src/arrays/programs_involving_data_sequences/medium/q7/MainTest.java
```diff
@@ -0,0 +1,61 @@
...
+                new StringLiteral("homecominghomecominghomecominghomecominghomecominghomecominghomecominghomecominghomecominghomecoming")
+        };
+    }
+
+    public void runMain() {
+        DupeDupeDupeDupe.main(new String[0]);
+    }
+
+    static Stream<Arguments> inputProvider() {
+        return Stream.of(
```
Hedgemon4, line 28: We should also have the test cases of a length 1 array and length 0 array. You can replace some of these other similar test cases with that.

src/arrays/programs_involving_data_sequences/medium/q7/MainTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void correctStringRepeaterMethod(String[] in) throws Throwable {
+        Object[][] arguments = {
+                {in, String[].class}
+        };
+        MethodTest m = new MethodTest(DupeDupeDupeDupe.class, "stringRepeater", arguments);
+        Object output = m.callMethod();
+        CustomAssertions._assertEquals(getSolution(in), output, "Your stringRepeater method does not return a string with enough repetitions of the longest word.");
```
Hedgemon4, line 46: Probably something about return a String with the correct number of repetitions of the longest word

Hedgemon4 (changes requested):

src/arrays/programs_involving_data_sequences/medium/q7/DupeDupeDupeDupe.java
```diff
@@ -0,0 +1,43 @@
+package arrays.programs_involving_data_sequences.medium.q7;
+
+/*
+Make a Java program that calls a method that takes an array of words as input. Go through all the words and find the
```
Hedgemon4, line 4: We never actually say that they are building a String, so I think we should rewrite this question to be more clear.

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #463, opened 2022-11-03
arrays-multidimensional_arrays-hard-q1

SamamthaH:
## PR Checklist
- [ ]  EOF newlines
---
## Website

### DIFF
--- a/src/arrays/multidimensional_arrays/hard/q1/AscendDimensions.java
+++ b/src/arrays/multidimensional_arrays/hard/q1/AscendDimensions.java
@@ -1,0 +1,55 @@
+package arrays.multidimensional_arrays.hard.q1;
+
+/*
+Make a program that asks the user to enter the number of rows and columns of a 2D integer array and then create that
+array in the program. The program also needs to ask the user for the values to initialize each element in that 2D array.
+The program then needs to print the elements of the array with each row on a separate line.
+
+Sample output:
+    Enter the number of rows:
+    2
+    Enter the number of columns:
+    3
+    Enter the number to store at row 1, column 1:
+    8
+    Enter the number to store at row 1, column 2:
+    7
+    Enter the number to store at row 1, column 3:
+    6
+    Enter the number to store at row 2, column 1:
+    5
+    Enter the number to store at row 2, column 2:
+    4
+    Enter the number to store at row 2, column 3:
+    3
+    8 7 6
+    5 4 3
+ */
+
+import java.util.Scanner;
+public class AscendDimensions
+{
+    public static void main( String [] args )
+    {
+        Scanner input = new Scanner( System.in );
+        System.out.println("Enter the number of rows: ");
+        int row = input.nextInt();
+        System.out.println("Enter the number of columns: ");
+        int col = input.nextInt();
+        int [][] nums = new int[row][col];
+        for( int i = 0; i < row; i++ )
+            for( int j = 0; j < col; j++ )
+            {
+                System.out.println("Enter the number to store at row " + ( i + 1 ) + ", column " + ( j + 1 ) + ": ");
+                nums[i][j] = input.nextInt();
+            }
+        for( int i = 0; i < row; i++ )
+        {
+            for( int j = 0; j < col; j++ )
+            {
+                System.out.print(nums[i][j] + " ");
+            }
+            System.out.println();
+        }
+    }
+}
--- a/src/arrays/multidimensional_arrays/hard/q1/AscendDimensions.java
+++ b/src/arrays/multidimensional_arrays/hard/q1/AscendDimensions.java
@@ -27,26 +27,22 @@ Sample output:
  */
 
 import java.util.Scanner;
-public class AscendDimensions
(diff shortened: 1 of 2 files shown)


SamamthaH:
I'm a bit worried about the last set of parameters failing when it hits the website, since I kept having to lower it because of a stack overflow.

Hedgemon4:
> I'm a bit worried about the last set of parameters failing when it hits the website, since I kept having to lower it because of a stack overflow.

As this is a parsons question, it should not be a problem if we need to limit the size of the 2D array.

src/arrays/multidimensional_arrays/hard/q1/MainTest.java
```diff
@@ -0,0 +1,112 @@
...
+            }
+        }
+    }
+
+    public static int[][] arrayBuilder(int rows, int columns) {
+        int[][] arr = new int[rows][columns];
+
+        for (int x = 0; x < rows; x++) {
+            for (int y = 0; y < columns; y++) {
+                arr[x][y] = (int) (Math.random() * 1000);
```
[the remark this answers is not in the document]
Hedgemon4: Sorry, I don't quite remember the syntax

src/arrays/multidimensional_arrays/hard/q1/MainTest.java
```diff
@@ -0,0 +1,112 @@
...
+                s.append(nums[x][y]);
+                s.append(" ");
+            }
+        }
+
+        runWithInput(s.toString(), clauseBuilder(rows, columns));
+
+        for (int x = 0; x < rows; x++) {
+            for (int y = 0; y < columns; y++) {
+                assertEquals(nums[x][y], Integer.parseInt(getItemByName("row" + x + "col" + y)), "The number on row " + x + " column " + y + " is incorrect.");
```
Hedgemon4, line 71: To follow out typical format, I think we should have the "Your program did not correctly ... " and then have the row and column identified in brackets " ... (the number on row x column y was incorrect.)

src/arrays/multidimensional_arrays/hard/q1/AscendDimensions.java
```diff
@@ -0,0 +1,51 @@
+package arrays.multidimensional_arrays.hard.q1;
+
+/*
+Make a program that asks the user to enter the number of rows and columns of a 2D integer array and then create that
```
Hedgemon4, line 4: Change question text to:
`Make a program that asks the user to enter the number of rows and columns of a 2D integer array and then create that array in the program. Also, the program should prompt the user for the values to initialize each element in that 2D array. Finally, print the elements of this array with each row on a separate line, with spaces between each number. Assume that the user will enter positive integers for the row and column numbers. `

src/arrays/multidimensional_arrays/hard/q1/MainTest.java
```diff
@@ -0,0 +1,120 @@
...
+                s.append(nums[x][y]);
+                s.append(" ");
+            }
+        }
+
+        runWithInput(s.toString(), clauseBuilder(rows, columns));
+
+        for (int x = 0; x < rows; x++) {
+            for (int y = 0; y < columns; y++) {
+                assertEquals(nums[x][y], Integer.parseInt(getItemByName("row" + x + "col" + y)), "Your program did not correctly print the number on row " + x + " column " + y + " is incorrect.");
```
Hedgemon4, line 81: I don't think you need to specify the row and column here

Hedgemon4: We can probably just say that it is not printed correctly

Hedgemon4 (changes requested):

src/arrays/multidimensional_arrays/hard/q1/MainTest.java
```diff
@@ -0,0 +1,120 @@
...
+                s.append(nums[x][y]);
+                s.append(" ");
+            }
+        }
+
+        runWithInput(s.toString(), clauseBuilder(rows, columns));
+
+        for (int x = 0; x < rows; x++) {
+            for (int y = 0; y < columns; y++) {
+                assertEquals(nums[x][y], Integer.parseInt(getItemByName("row" + x + "col" + y)), "Your program did not correctly print the array.");
```
Hedgemon4, line 81:
```suggestion
                assertEquals(nums[x][y], Integer.parseInt(getItemByName("row" + x + "col" + y)), "Your program did not correctly print the input array.");
```

src/arrays/multidimensional_arrays/hard/q1/AscendDimensions.java
```diff
@@ -0,0 +1,52 @@
...
+array in the program. Also, the program should prompt the user for the values to initialize each element in that 2D array.
```
Hedgemon4, line 5:
```suggestion
array in the program. Then, the program should prompt the user for the values to initialize each element in that 2D array.
```

Hedgemon4 (approved):

PR #468, opened 2022-11-08
arrays-programs_involving_multidimensional_data-hard-q1

SamamthaH:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/arrays/programs_involving_multidimensional_data/hard/q1/Sums.java
```diff
@@ -0,0 +1,60 @@
+package arrays.programs_involving_multidimensional_data.hard.q1;
+
+/*
+Write a Java program called Sums. It should have a main method which takes user input to create a 2D array and calls
```
Hedgemon4, line 4: Change question text to:
`Write a Java program called Sums. It should have a main method which takes user input to create a 2D integer array and calls another method called sums. The sums method should take a 2D integer array as a parameter and return a 1D integer array which is the sum of each row in the 2D array. The main method should then print the 1D array.`

src/arrays/programs_involving_multidimensional_data/hard/q1/MainTest.java
```diff
@@ -0,0 +1,164 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MainTest extends BaseTest {
+    // Java
+
+    public Clause[] testSentence() {
+        TestOption.isInputTest = true;
+        TestOption.defaultInput = "3 2 4 32 4 2 12 32";
+        return new Clause[]{
+                new StringLiteral("Enter the number of rows:"),
```
Hedgemon4, line 29: I think instead of asking for input for everything, you should change it to be constant (like 4 by 4), and then we can vary that on the website

Hedgemon4 (changes requested):

src/arrays/programs_involving_multidimensional_data/hard/q1/MainTest.java
```diff
@@ -0,0 +1,156 @@
...
+        int[] sums = new int[size];
+
+        for (int x = 0; x < size; x++) {
+            for (int y = 0; y < size; y++) {
+                sums[x] += input[x][y];
+            }
+        }
+
+        for (int x = 0; x < size; x++) {
+            assertEquals(sums[x], Integer.parseInt(getItemByName(String.valueOf(x))), "Your program does not print the correct sum of the elements in a row.");
```
Hedgemon4, line 118:
```suggestion
            assertEquals(sums[x], Integer.parseInt(getItemByName(String.valueOf(x))), "Your program does not print the correct sum of the elements in each row of the array.");
```

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #472, opened 2022-11-09
arrays-programs_involving_multidimensional_data-easy-q1

SamamthaH:
## PR Checklist
- [x]  EOF newlines
---
## Website

### DIFF
--- a/src/arrays/programs_involving_multidimensional_data/easy/q1/AllSameSum.java
+++ b/src/arrays/programs_involving_multidimensional_data/easy/q1/AllSameSum.java
@@ -5,7 +5,7 @@ Write a Java program that initialises a 4 by 4 multidimensional array of integer
 prints true if all the rows have the same sum and false otherwise.
 
 Sample Output:
-    Enter a 4 by 4 multidimentional array in 4 lines:
+    Enter a 4 by 4 multidimensional array in 4 lines:
     4 5 3 2
     7 8 -1 0
     10 1 1 2
@@ -20,7 +20,7 @@ public class AllSameSum {
         Scanner input = new Scanner(System.in);
         int[][] arr = new int[4][4];
 
-        System.out.println("Enter a 4 by 4 multidimentional array in 4 lines:");
+        System.out.println("Enter a 4 by 4 multidimensional array in 4 lines:");
         for (int x = 0; x < 4; x++)
             for (int y = 0; y < 4; y++)
                 arr[x][y] = input.nextInt();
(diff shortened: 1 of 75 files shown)


src/arrays/programs_involving_multidimensional_data/easy/q1/MainTest.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.stream.Stream;
+
+public class MainTest extends BaseTest {
+    // Parsons
+
+    static final int size = 4;
+
+    public Clause[] testSentence() {
+        TestOption.isInputTest = true;
+        TestOption.defaultInput = "1 2 3 4 4 3 2 1 2 4 1 3 3 4 2 1";
```
Hedgemon4, line 23: We should probably change this to be generated input so we can vary the array size easier

src/arrays/programs_involving_multidimensional_data/easy/q1/MainTest.java
```diff
@@ -0,0 +1,103 @@
...
+    public void runMain() {
+        AllSameSum.main(new String[0]);
+    }
+
+    static Stream<int[][]> inputProvider() {
+        int[][] one = new int[size][size];
+        for (int x = 0; x < size; x++) {
+            one[x] = ArrayUtil.generateRandomArray(0, 10, size);
+        }
+        int[][] two = new int[size][size];
```
Hedgemon4, line 36: Are any of these going to actually be magic squares?

src/arrays/programs_involving_multidimensional_data/easy/q1/AllSameSum.java
```diff
@@ -0,0 +1,33 @@
...
+A magic square is a box of numbers where each row and column has the same sum. Write a Java program that initialises a
+4 by 4 multidimensional array of integers with values given by the user, and prints true if all the rows have the same
+sum and false otherwise.
+ */
+
+import java.util.Scanner;
+
+public class AllSameSum {
+    public static void main(String[] args) {
+        Scanner input = new Scanner(System.in);
```
Hedgemon4, line 13: This program only checks if each row has the same sum, not if each row and column has the same sum (as the question specifies)

Hedgemon4: So we either need to change this so that it checks both, or to change the question to just check the rows

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #476, opened 2022-11-11
arrays-programs_involving_data_sequences-hard-q6

KaiKhademi:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/arrays/programs_involving_data_sequences/hard/q6/FibBeat.java
```diff
@@ -0,0 +1,42 @@
+package arrays.programs_involving_data_sequences.hard.q6;
+
+/*
+Recall that the Fibonacci sequence starts with F0 = 0, F1=1, and then all future values are defined by the previous
```
SamamthaH, line 4: Change question text to:
"Recall that the Fibonacci sequence starts with F0 = 0, F1=1, and then all future values are defined by the previous values using the formula: Fn= Fn-1+Fn-2. Write a Java program that asks the user for an integer n and calls a method called fibonacciMaker which takes n as an input and generates an array of the first n numbers in the sequence. The method should return the array and the main method print each element separated by a space. Because the Fibonacci sequence starts at F0 the size of the array would be n+1. For example, if n is 5, then the method returns the array [0,1,1,2,3,5]. You can assume that n is always positive. Your method should be general enough to work with any size of n."

src/arrays/programs_involving_data_sequences/hard/q6/MainTest.java
```diff
@@ -0,0 +1,85 @@
...
+    public Clause[] testSentence() {
+        TestOption.isInputTest = true;
+        TestOption.defaultInput = "5";
+        return new Clause[]{
+                new StringLiteral("Enter a positive integer n:"),
+                new Optional(new StringLiteral(" ")),
+                new NewLine(),
+                new StringLiteral("The Fibonacci sequence of "),
+                new IntegerLiteral(),
+                new StringLiteral(" is:"),
```
SamamthaH, line 31: Add another optional here.

src/arrays/programs_involving_data_sequences/hard/q6/MainTest.java
```diff
@@ -0,0 +1,85 @@
...
+                new NewLine(),
+                new PlaceHolder(),
+                new Optional(new StringLiteral(" "))
```
SamamthaH, line 34: Move this inside the placeholders

src/arrays/programs_involving_data_sequences/hard/q6/MainTest.java
```diff
@@ -0,0 +1,85 @@
...
+        return out;
+    }
+
+    static Stream<Arguments> InputProvider() {
+        return Stream.of(
+                Arguments.of(1),
+                Arguments.of(10),
+                Arguments.of(100),
+                Arguments.of(10000)
+
```
SamamthaH, line 59: get ride of extra spaces

src/arrays/programs_involving_data_sequences/hard/q6/MainTest.java
```diff
@@ -0,0 +1,85 @@
...
+        out[0] = 0;
+        if (n >= 1)
+            out[1] = 1;
+        if (n >= 2)
+            for (int i = 2; i <= n; i++)
+                out[i] = out[i - 1] + out[i - 2];
+        return out;
+    }
+
+    static Stream<Arguments> InputProvider() {
```
SamamthaH, line 53:
```suggestion
    static Stream<Arguments> inputProvider() {
```

src/arrays/programs_involving_data_sequences/hard/q6/MainTest.java
```diff
@@ -0,0 +1,85 @@
...
+                Arguments.of(10),
+                Arguments.of(100),
+                Arguments.of(10000)
+
+
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("InputProvider")
```
SamamthaH, line 65:
```suggestion
    @MethodSource("inputProvider")
```

src/arrays/programs_involving_data_sequences/hard/q6/MainTest.java
```diff
@@ -0,0 +1,85 @@
...
+    void printCorrectOutput(int n) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not print the correct fibonacci sequence.";
+        runWithInput(Integer.toString(n), new Clause[]{
+                new StringLiteral(arrayToInput(fib(n)))
+        });
+    }
+
+    @ParameterizedTest
+    @MethodSource("InputProvider")
```
SamamthaH, line 74:
```suggestion
    @MethodSource("inputProvider")
```

src/arrays/programs_involving_data_sequences/hard/q6/MainTest.java
```diff
@@ -0,0 +1,85 @@
...
+    void correctFibonacciMaker(int n) throws Throwable {
+        Object[][] arguments = {
+                {n, int.class},
+        };
+        MethodTest m = new MethodTest(FibBeat.class, "fibonacciMaker", arguments);
+        m.setIncorrectMethodStructureErrorMessage("Your fibonacciMaker method does not calculate the correct fibonacci sequence up to nth element.");
+        Object output = m.callMethod();
+        assertArrayEquals((int[]) output, fib(n));
```
SamamthaH, line 82: Switch to CustomAssertions, move the expected value to be in front of the output, and move the incorrect structure error to the assert error message.

SamamthaH (changes requested):

SamamthaH (approved):

PR #477, opened 2022-11-11
arrays-programs_involving_data_sequences-hard-q7

KaiKhademi:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/arrays/programs_involving_data_sequences/hard/q7/SquareUp.java
```diff
@@ -0,0 +1,37 @@
+package arrays.programs_involving_data_sequences.hard.q7;
+
+/*
+Write a Java program that calls a method called doubler that takes as input an integer array, and squares each element
```
SamamthaH, line 4: Change question text to:
"Write a Java program that asks the user to input 10 integers and stores them in an array. Then it should pass them to a method called doubler takes an input of an integer array and stores the square of each element in a new array which it then returns. After calling the method, the main method should print the contents of the new array separated by a space. Your method should be general enough to work with any size of array."

src/arrays/programs_involving_data_sequences/hard/q7/MainTest.java
```diff
@@ -0,0 +1,96 @@
...
+    public static int n = 10;
+
+    public Clause[] testSentence() {
+        TestOption.isInputTest = true;
+        TestOption.defaultInput = arrayToInput(generateRandomArray(-10, 10, n));
+        return new Clause[]{
+                new StringLiteral("Enter an array of " + n + " integers to be squared:"),
+                new Optional(new StringLiteral(" ")),
+                new NewLine(),
+                new StringLiteral("The squared array is:"),
```
SamamthaH, line 31: Add an optional space before the NewLine

src/arrays/programs_involving_data_sequences/hard/q7/MainTest.java
```diff
@@ -0,0 +1,96 @@
...
+                new NewLine(),
+                new PlaceHolder(),
+                new Optional(new StringLiteral(" "))
```
SamamthaH, line 34:
```suggestion
```

SamamthaH: PlaceHolder will capture any remaining data so this will never be used.

src/arrays/programs_involving_data_sequences/hard/q7/MainTest.java
```diff
@@ -0,0 +1,96 @@
...
+    public static int[] sq(int[] arr) {
+        int[] input = arr.clone();
+        for (int i = 0; i < input.length; i++)
+            input[i] = (int) Math.pow(input[i], 2);
+        return input;
+    }
+
+    static Stream<Arguments> mainInputProvider() {
+        return Stream.of(
+                Arguments.of(generateRandomArray(-10000, 10000, n)),
```
SamamthaH, line 51: Vary the range of these arrays.

src/arrays/programs_involving_data_sequences/hard/q7/MainTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("mainInputProvider")
+    void printCorrectOutput(int[] a) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not print the correct sequence of squared array.";
+        runWithInput(arrayToInput(a), new Clause[]{
+                new StringLiteral(arrayToInput(sq(a)))
```
SamamthaH, line 68: do you need to add the optional here instead?

src/arrays/programs_involving_data_sequences/hard/q7/MainTest.java
```diff
@@ -0,0 +1,96 @@
...
+        });
+    }
+
+    static Stream<Arguments> methodInputProvider() {
+        return Stream.of(
+                Arguments.of(generateRandomArray(-10000, 10000, 10000)),
```
SamamthaH, line 74: There should be some hardcoded arrays in here, and the generated ones should have varying lengths and ranges.

src/arrays/programs_involving_data_sequences/hard/q7/MainTest.java
```diff
@@ -0,0 +1,96 @@
...
+    @ParameterizedTest
+    @MethodSource("methodInputProvider")
+    void correctFibonacciMaker(int[] input) throws Throwable {
+        Object[][] arguments = {
+                {input, int[].class},
+        };
+        MethodTest m = new MethodTest(SquareUp.class, "doubler", arguments);
+        m.setIncorrectMethodStructureErrorMessage("Your doubler method does not return the correct array of squared integers.");
+        Object output = m.callMethod();
+        assertArrayEquals((int[]) output, sq(input));
```
SamamthaH, line 93: Please calculate the expected output before you call the method in case the student does not create a new array and passes back the original one instead. Also, move the expected value before the output, use CustomAssertions, and make you incorrect structure error a test error.

SamamthaH (changes requested):

src/arrays/programs_involving_data_sequences/hard/q7/MainTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+    public void runMain() {
+        SquareUp.main(new String[0]);
+    }
+
+    public static int[] sq(int[] arr) {
+        int[] input = arr.clone();
+        for (int i = 0; i < input.length; i++)
+            input[i] = (int) Math.pow(input[i], 2);
+        return input;
```
SamamthaH, line 47: To change this to be different than the solution code, you can have input[i]*=input[i] instead of Math.pow

SamamthaH (changes requested):

src/arrays/programs_involving_data_sequences/hard/q7/MainTest.java
```diff
@@ -0,0 +1,102 @@
...
+                Arguments.of(replicateArray(10000, n))
+
+
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("mainInputProvider")
+    void printCorrectOutput(int[] a) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not print the correct sequence of squared array.";
```
SamamthaH, line 67:
```suggestion
        TestOption.incorrectStructureErrorMessage = "Your program does not correctly print the squared array.";
```

SamamthaH (changes requested):

SamamthaH (approved):

PR #478, opened 2022-11-11
arrays-programs_involving_data_sequences-hard-q8

KaiKhademi:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/arrays/programs_involving_data_sequences/hard/q8/MainTest.java
```diff
@@ -0,0 +1,96 @@
...
+    @ParameterizedTest
+    @MethodSource("methodInputProvider")
+    void correctFibonacciMaker(int[] input) throws Throwable {
+        Object[][] arguments = {
+                {input, int[].class},
+        };
+        MethodTest m = new MethodTest(MultItUp.class, "productMaker", arguments);
+        m.setIncorrectMethodStructureErrorMessage("Your productMaker method does not return the correct multiplied array");
+        Object output = m.callMethod();
+        assertArrayEquals((int[]) output, mul(input));
```
SamamthaH, line 94: Currently, there is no way to fail this test. Since the java method alters the array you pass it instead of making a new one, you are grabbing the same reference in the output as is in the input so a comparison between the two will never fail.

SamamthaH: Also, you should be using CustomAssertions which would not require you to cast output and include an error message in the test.

src/arrays/programs_involving_data_sequences/hard/q8/MainTest.java
```diff
@@ -0,0 +1,96 @@
...
+        TestOption.isInputTest = true;
+        TestOption.defaultInput = arrayToInput(generateRandomArray(-10, 10, n));
+        return new Clause[]{
+                new StringLiteral("Enter an array of " + n + " integers:"),
+                new Optional(new StringLiteral(" ")),
+                new NewLine(),
+                new StringLiteral("The result of multiplied array is:"),
+                new NewLine(),
+                new PlaceHolder(),
+                new Optional(new StringLiteral(" "))
```
SamamthaH, line 34:
```suggestion
```

SamamthaH: The PlaceHolder will capture all the input after the NewLine so this will never capture anything

src/arrays/programs_involving_data_sequences/hard/q8/MultItUp.java
```diff
@@ -0,0 +1,39 @@
...
+        System.out.println("Enter an array of 10 integers:");
+        for (int i = 0; i < 10; i++)
+            arr[i] = in.nextInt();
+        int[] result = productMaker(arr);
+        System.out.println("The result of multiplied array is:");
+        for (int i = 0; i < result.length; i++)
+            System.out.print(result[i] + " ");
+    }
+
+    public static int[] productMaker(int[] input) {
```
SamamthaH, line 32:
```suggestion
    public static void productMaker(int[] input) {
```

src/arrays/programs_involving_data_sequences/hard/q8/MultItUp.java
```diff
@@ -0,0 +1,39 @@
...
+        for (int i = 0; i < input.length; i++) {
+            int value = input[i] * i;
+            input[i] = value;
+        }
+        return input;
```
SamamthaH, line 37:
```suggestion
```

src/arrays/programs_involving_data_sequences/hard/q8/MultItUp.java
```diff
@@ -0,0 +1,39 @@
+package arrays.programs_involving_data_sequences.hard.q8;
+
+/*
+Write a Java program that calls a method called productMaker that takes as input an array of integers. The method
```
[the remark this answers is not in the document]
SamamthaH: You will have to change your test to reflect that the method no longer has a return value.

src/arrays/programs_involving_data_sequences/hard/q8/MainTest.java
```diff
@@ -0,0 +1,96 @@
...
+        for (int i = 0; i < input.length; i++) {
+            int value = input[i] * i;
+            input[i] = value;
+        }
+        return input;
+    }
+
+    static Stream<Arguments> mainInputProvider() {
+        return Stream.of(
+                Arguments.of(generateRandomArray(-10000, 10000, n)),
```
SamamthaH, line 52: The ranges of these first three could be varied

src/arrays/programs_involving_data_sequences/hard/q8/MainTest.java
```diff
@@ -0,0 +1,96 @@
...
+    void printCorrectOutput(int[] a) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not print the correct multiplied array.";
+        runWithInput(arrayToInput(a), new Clause[]{
+                new StringLiteral(arrayToInput(mul(a)))
+        });
+    }
+
+    static Stream<Arguments> methodInputProvider() {
+        return Stream.of(
+                Arguments.of(generateRandomArray(-10000, 10000, 10000)),
```
SamamthaH, line 75: Add in some hardcoded arrays and vary the length and range of the generated ones.

src/arrays/programs_involving_data_sequences/hard/q8/MainTest.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @ParameterizedTest
+    @MethodSource("methodInputProvider")
+    void correctFibonacciMaker(int[] input) throws Throwable {
+        Object[][] arguments = {
+                {input, int[].class},
+        };
+        MethodTest m = new MethodTest(MultItUp.class, "productMaker", arguments);
+        m.setIncorrectMethodStructureErrorMessage("Your productMaker method does not return the correct multiplied array");
```
SamamthaH, line 92: Since the method does not print anything, this will never trigger.

SamamthaH (changes requested):

SamamthaH (changes requested):
Please finish implementing prior requested changes, or give a reason as to why you don't think you should.

src/arrays/programs_involving_data_sequences/hard/q8/MultItUp.java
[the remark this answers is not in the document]
SamamthaH: Still need to change the question text

src/arrays/programs_involving_data_sequences/hard/q8/MainTest.java
```diff
@@ -0,0 +1,102 @@
...
+                Arguments.of(replicateArray(10000, n))
+
+
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("mainInputProvider")
+    void printsCorrectOutput(int[] a) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not print the correct multiplied array.";
```
SamamthaH, line 67:
```suggestion
        TestOption.incorrectStructureErrorMessage = "Your program does not correctly print the multiplied array.";
```

src/arrays/programs_involving_data_sequences/hard/q8/MainTest.java
```diff
@@ -0,0 +1,102 @@
...
+    @ParameterizedTest
+    @MethodSource("methodInputProvider")
+    void correctProductMakerr(int[] input) throws Throwable {
+        int[] ans = mul(input);
+        Object[][] arguments = {
+                {input, int[].class},
+        };
+        MethodTest m = new MethodTest(MultItUp.class, "productMaker", arguments);
+        m.callMethod();
+        CustomAssertions._assertArrayEquals(ans, input, "Your productMaker method does not return the correct multiplied array");
```
SamamthaH, line 100:
```suggestion
        CustomAssertions._assertArrayEquals(ans, input, "Your productMaker method does not return a correctly multiplied array");
```

SamamthaH (changes requested):

src/arrays/programs_involving_data_sequences/hard/q8/MultItUp.java
```diff
@@ -0,0 +1,38 @@
...
+Write a Java program that asks the user to enter 10 integers, stores them in an array, and passes that array to a method
+called productMaker. This method should take an array of integers and multiply each element by its index. The method
+should have no return value as the changes to the array will carry back over to the main method. The main method should
+then print the result with a space between each element. The productMaker method should be able to work with an array of
+any size. Ex. the array [1,4,6] will become [0,4,12]
+
+Sample Output:
+    Enter an array of 10 integers:
+    1 2 3 4 5 6 7 8 9 10
+    The result of multiplied array is::
```
SamamthaH, line 13:
```suggestion
    The result of multiplied array is:
```

src/arrays/programs_involving_data_sequences/hard/q8/MultItUp.java
```diff
@@ -0,0 +1,38 @@
+package arrays.programs_involving_data_sequences.hard.q8;
+
+/*
+Write a Java program that asks the user to enter 10 integers, stores them in an array, and passes that array to a method
```
SamamthaH, line 4: Number of inputs needs to be higher, also could you check if moving the 
`for (int i = 0; i < 10; i++)
            arr[i] = in.nextInt();
        productMaker(arr);`
after the "The result of multiplied array is:" statement to check if the test still approves it. If it does, something about it needs to change since they can just for loop catch and print the values without storing them.

src/arrays/programs_involving_data_sequences/hard/q8/MainTest.java
```diff
@@ -0,0 +1,102 @@
...
+    public static int n = 10;
+
+    public Clause[] testSentence() {
+        TestOption.isInputTest = true;
+        TestOption.defaultInput = arrayToInput(generateRandomArray(-10, 10, n));
+        return new Clause[]{
+                new StringLiteral("Enter an array of " + n + " integers:"),
+                new Optional(new StringLiteral(" ")),
+                new NewLine(),
+                new StringLiteral("The result of multiplied array is:"),
```
SamamthaH, line 31: add optional in here

SamamthaH (changes requested):

src/arrays/programs_involving_data_sequences/hard/q8/MultItUp.java
KaiKhademi: They won't be able to because we are using the method that takes an array.

src/arrays/programs_involving_data_sequences/hard/q8/MainTest.java
```diff
@@ -0,0 +1,102 @@
...
+    @ParameterizedTest
+    @MethodSource("methodInputProvider")
+    void correctProductMakerr(int[] input) throws Throwable {
+        int[] ans = mul(input);
+        Object[][] arguments = {
+                {input, int[].class},
+        };
+        MethodTest m = new MethodTest(MultItUp.class, "productMaker", arguments);
+        m.callMethod();
+        CustomAssertions._assertArrayEquals(ans, input, "Your productMaker method does not return a correctly multiplied array");
```
SamamthaH, line 100:
```suggestion
        CustomAssertions._assertArrayEquals(ans, input, "Your productMaker method does not correctly multiply the array parameter.");
```

SamamthaH (changes requested):

SamamthaH (approved):

PR #479, opened 2022-11-13
arrays-programs_involving_multidimensional_data-medium-q1

SamamthaH:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/arrays/programs_involving_multidimensional_data/medium/q1/Identical.java
```diff
@@ -0,0 +1,77 @@
+package arrays.programs_involving_multidimensional_data.medium.q1;
+
+/*
+Write a Java program that initializes two integers arrays of size 5 by 5 and takes user input values for them. Then,
```
Hedgemon4, line 4: Change question text to:
`Write a Java program that initializes two integer arrays of size 5 by 5 and prompts the user to input values for them. Then, check to see if each element in the first array matches the corresponding element in the second array. If any pair of elements are not the same, print "different", otherwise, do nothing. For example, if arr1[x][y] = 5, and arr2[x][y] = 6, you would print different. `

src/arrays/programs_involving_multidimensional_data/medium/q1/MainTest.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    public void runMain() {
+        Identical.main(new String[0]);
+    }
+
+    static Stream<Arguments> inputProvider() {
+        int[][] a1 = new int[size][size];
+        int[][] b1 = new int[size][size];
+
```
Hedgemon4, line 35: You should add a test case where the numbers aren't random. You could try doing something with ascending arrays or some other arrayUtil function

Hedgemon4 (changes requested):

src/arrays/programs_involving_multidimensional_data/medium/q1/MainTest.java
SamamthaH: that's what i was trying to do with
a3[x] = ArrayUtil.generateRandomArray(1, 2, size);
b3[x] = ArrayUtil.generateRandomArray(1, 2, size);
but after reviewing Kai's prs i remembered that the replicate method exists in array util so I can switch them to that

src/arrays/programs_involving_multidimensional_data/medium/q1/Identical.java
```diff
@@ -0,0 +1,78 @@
...
+int[][] b = new int[][];
+int[] a = new int[5];
+int[] b = new int[5];
+public static void main (String args)
+int[][] b = new int[5][5]
+ */
+
+import java.util.Scanner;
+
+public class Identical {
```
Hedgemon4, line 51: I feel like the Scanner is kinda overkill for this because it is Parsons with distractors. I would just simplify this down to giving them two arrays and then focus more on the comparison portion of the question.

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #480, opened 2022-11-14
Created ObjectTest for testing Objects

Hedgemon4:


src/global/ObjectTest.java
```diff
@@ -0,0 +1,485 @@
...
+                if (namesSet.contains(clause.getName())) {
+                    _fail("There is an issue with the test definition. Please contact an administrator.",
+                            "The name " + clause.getName() + " is already in use. Make sure all names are unique");
+                }
+                namesSet.add(clause.getName());
+            }
+            try {
+                clause.validateParams();
+            } catch (InvalidClauseException e) {
+                throw new InvalidClauseException("The parameter(s) for the " + methodName + " method clause at index " + i + " are invalid. " + e.getMessage());
```
keyvankhademi, line 476: What will happen when we throw an error without `_fail`? What does the user see?

src/global/ObjectTest.java
```diff
@@ -0,0 +1,485 @@
...
+            to check method console output using clauses. Note that random clauses and the placeHolder clause is not
+            currently supported for methods.
+         */
+
+        String output = methodOutput.toString();
+
+        if (methodTestSentence == null) {
+            assertEquals("", output,
+                    String.join(
+                            "",
```
keyvankhademi, line 423: Why are we having an empty string at the beginning?

keyvankhademi: oh that's what it joins it with.

src/global/ObjectTest.java
```diff
@@ -0,0 +1,485 @@
...
+                                    "class does not have the correct modifiers."));
+                }
+            }
+            ByteArrayOutputStream methodOutput = new ByteArrayOutputStream();
+            System.setOut(new PrintStream(methodOutput));
+            Object output = objectMethodInvoke.invoke(object, args);
+            methodOutputFollowsCorrectStructure(methodName, methodTestSentence, incorrectMethodStructureErrorMessage, methodOutput);
+            return output;
+        } catch (NoSuchMethodException e) {
+            fail(String.join("", "Your ", objectClass.getSimpleName(),
```
keyvankhademi, line 377: Why do we have both `fail` and `_fail`? Whats the difference?

keyvankhademi (dismissed):
This is great. Seems to be error-proof too.

src/global/ObjectTest.java
Hedgemon4: _fail allows us to use two messages. The first is the error message the user will see, and the second logs a developer message to the console which tells them if they made a mistake with something, or gives them more description of the error.

src/global/ObjectTest.java
Hedgemon4: I think they will see this, but this should be changed to fail.

Hedgemon4: This is actually a developer error if the clause is not defined correctly, so the student should never see this error. However, I will change it to use _fail anyways.

Hedgemon4: Unless there is an error with our test, but they should be tested as they are put on the website to make sure this doesn't happen.

src/global/ObjectTest.java
```diff
@@ -0,0 +1,506 @@
...
+    }
+
+    public Object createArray(int length) {
+        if (length < 0)
+            _fail("Error with test definition. Please contact a test administrator.",
+                    "Must pass a positive length to initialize an array.");
+        return Array.newInstance(this.objectClass, length);
+    }
+
+    public Object createArray(int length, Object[][][] arguments) throws Throwable {
```
keyvankhademi, line 497: Let's add a method that creates identical elements with the same input as well.

keyvankhademi (approved):
Well done.

PR #481, opened 2022-11-15
arrays-programs_involving_multidimensional_data-hard-q2

SamamthaH:
## PR Checklist
- [ ]  EOF newlines
---
## Website

### DIFF
--- a/src/arrays/programs_involving_multidimensional_data/hard/q2/Swapped.java
+++ b/src/arrays/programs_involving_multidimensional_data/hard/q2/Swapped.java
@@ -20,29 +20,27 @@ Sample output:
     55 50 19 59 40
  */
 
-public class Swapped
-{
-    public static void main( String[] args )
-    {
+public class Swapped {
+    public static void main(String[] args) {
         int[][] arr = new int[5][5];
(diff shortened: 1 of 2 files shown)


src/arrays/programs_involving_multidimensional_data/hard/q2/Swapped.java
```diff
@@ -0,0 +1,49 @@
+package arrays.programs_involving_multidimensional_data.hard.q2;
+
+/*
+Write a Java program called Swapped, in the main method initialize a 5 by 5 array and fill it with random numbers
```
Hedgemon4, line 4: Change question text to:
`Write a Java program called Swapped and in the main method initialize a 5 by 5 array and fill it with random numbers between 0 and 100 (inclusive). Print out the array with spaces between each number and a new line between each row. Then, swap the first and last item in each row of the array and print the resulting array in the same format as the first one. Note that there should be an empty line separating the two printed arrays. `

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #486, opened 2022-11-16
arrays-multidimensional_arrays-hard-q4

KaiKhademi:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/arrays/multidimensional_arrays/hard/q4/AlterTable.java
```diff
@@ -0,0 +1,22 @@
+package arrays.multidimensional_arrays.hard.q4;
+
+/*
+Make a program that defines a 2D array and goes through each element in it by traversing each row's elements one at a
```
SamamthaH, line 4: Change question text to:
"Make a program that defines a multidimensional array and iterates over each row of the array, printing out every other element starting from the first element of the first row. For example, if you have the following 2D array {{1,2,3}, {4, 5 ,6}, {7,8,9} }, the program will print 1, 3, 5, 7, 9."

src/arrays/multidimensional_arrays/hard/q4/MainTest.java
```diff
@@ -0,0 +1,34 @@
+package arrays.multidimensional_arrays.hard.q4;
+
+import global.BaseTest;
+import global.variables.Clause;
+import global.variables.clauses.*;
+
+public class MainTest extends BaseTest {
+    // Parsons
+
+    public static int[][] arr = {{4, 5, 6}, {66, 54, 21}, {609, 998, 543}, {1111, 2342, 4583}};
```
SamamthaH, line 10:
```suggestion
```

src/arrays/multidimensional_arrays/hard/q4/MainTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public Clause[] testSentence() {
+        return new Clause[]{
+                new StringLiteral(answerFor(arr))
```
SamamthaH, line 14: The answer here should be hardcoded in as it will never change.

src/arrays/multidimensional_arrays/hard/q4/MainTest.java
```diff
@@ -0,0 +1,34 @@
...
+        };
+    }
+
+    public void runMain() {
+        AlterTable.main(new String[0]);
+    }
+
+    public static String answerFor(int[][] nums) {
```
SamamthaH, line 22: This helper method is not required for the question.

SamamthaH (changes requested):

src/arrays/multidimensional_arrays/hard/q4/AlterTable.java
SamamthaH: The question text was not changed

SamamthaH (changes requested):

SamamthaH (approved):

PR #488, opened 2022-12-05
arrays-multidimensional_arrays-hard-q6

KaiKhademi:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/arrays/multidimensional_arrays/hard/q6/MainTest.java
```diff
@@ -0,0 +1,49 @@
+package arrays.multidimensional_arrays.hard.q6;
+
+import global.BaseTest;
+import global.variables.Clause;
+import global.variables.clauses.NewLine;
+import global.variables.clauses.StringLiteral;
+import global.variables.wrappers.Optional;
+
+public class MainTest extends BaseTest {
```
SamamthaH, line 9: Change test according to question changes

SamamthaH (changes requested):

src/arrays/multidimensional_arrays/hard/q6/OneAndOneMakeTwoDArrays.java
```diff
@@ -0,0 +1,42 @@
+package arrays.multidimensional_arrays.hard.q6;
+
+/*
+Make a program that defines two 2D integer arrays and uses a method called addMatrices to add them together to create a
```
SamamthaH, line 4: Realized that the way I asked for the new question to be set up leaves it very unclear where the error is supposed to be printed so change text to:
"Make a program that defines two 2D integer arrays and uses a method called addMatrices to add them together to create a new 2D array. The main method should define the two initial arrays and pass them to the addMatrices method. The method should check that the two arrays have the same dimensions, otherwise returning null. To add two arrays you take the elements at the same index, add them together, and place the new number at the same index in the new array. For the sake of simplicity, assume that the rows in an array have a consistent length. In the main method, if the method returned null, print an error message and otherwise print the newly created 2D array."

src/arrays/multidimensional_arrays/hard/q6/OneAndOneMakeTwoDArrays.java
```diff
@@ -0,0 +1,42 @@
...
+        int [][] c = addMatrices(a, b);
+        if(c != null){
+            for(int i = 0; i < c.length; i ++){
+                for(int j = 0; j < c[0].length; j ++)
+                    System.out.print(c[i][j] + " ");
+                System.out.println();
+            }
+        }
+        else
+            System.out.println("Can't add because arrays of different sizes!");
```
SamamthaH, line 25:
```suggestion
            System.out.println("Can't add the arrays because they are different sizes!");
```

SamamthaH: This will require changes to the test as well

src/arrays/multidimensional_arrays/hard/q6/MainTest.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+        else{
+            Clause[] c = new Clause[3 * a.length];
+            int[][] ans = answerFor(a, b);
+            for(int i = 0; i < a.length; i ++){
+                String st = "";
+                for(int j = 0; j < a[0].length; j ++){
+                    st += ans[i][j] + " ";
+                }
+                c[3 * i] = new StringLiteral(st.trim());
```
SamamthaH, line 38: Since this is Parsons, the students won't have the option of using .trim() and therefore the added " " at the end of a line will always be there

src/arrays/multidimensional_arrays/hard/q6/MainTest.java
```diff
@@ -0,0 +1,102 @@
...
+            return null;
+        int[][] ans = new int[x.length][x[0].length];
+        for(int i = 0; i < x.length; i ++){
+            for(int j = 0; j < x[0].length; j ++)
+                ans[i][j] = x[i][j] + y[i][j];
+        }
+        return ans;
+    }
+
+    static Stream<Arguments> inputProvider() {
```
SamamthaH, line 61: You can hardcode the answers to this instead of relying on a method

src/arrays/multidimensional_arrays/hard/q6/MainTest.java
```diff
@@ -0,0 +1,102 @@
...
+        Object[][] arguments = {
+                {a, int[][].class},
+                {b, int[][].class}
+        };
+        MethodTest m = new MethodTest(OneAndOneMakeTwoDArrays.class, "addMatrices", arguments);
+        Object output = m.callMethod();
+        if(output != null)
+            CustomAssertions._assertArrayEquals(answerFor(a, b), output, "Your addMatrices method does not correctly add two matrices.");
+        else
+            assertTrue(answerFor(a, b) == null, "Your addMatrices method does not give the correct answer.");
```
SamamthaH, line 100:
```suggestion
            assertTrue(answerFor(a, b) == null, "Your addMatrices method does not correctly identify invalid input.");
```

SamamthaH (changes requested):

src/arrays/multidimensional_arrays/hard/q6/MainTest.java
KaiKhademi: I will still need the method for testSentence for testing the output of program, it looked more neat to do the same here is there any emphasis on hardcoding the solution whenever possible?

SamamthaH: It's more of a thing because we try not to rely on machine code to generate our answers when we can write them out by hand (keyboard) and thus be sure that errors within the code aren't perpetuated. It doesn't interrupt any generated cases in inputProvider, so hard coding in the solutions for hard coded test cases is preferred.

src/arrays/multidimensional_arrays/hard/q6/MainTest.java
```diff
@@ -0,0 +1,101 @@
...
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    public void correctAddMatrices(int[][] a, int[][] b) throws Throwable {
+        Object[][] arguments = {
+                {a, int[][].class},
+                {b, int[][].class}
+        };
+        MethodTest m = new MethodTest(OneAndOneMakeTwoDArrays.class, "addMatrices", arguments);
+        Object output = m.callMethod();
+        if(output != null)
```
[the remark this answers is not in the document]
KaiKhademi: Did you meant the opposite here? because it is already dependent on what hey give back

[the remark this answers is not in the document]
KaiKhademi: Okay now I'll change it tell me if it's wrong

src/arrays/multidimensional_arrays/hard/q6/OneAndOneMakeTwoDArrays.java
```diff
@@ -0,0 +1,40 @@
+package arrays.multidimensional_arrays.hard.q6;
+
+/*
+Make a program that defines two 2D integer arrays and uses a method called addMatrices to add them together to create a
+new 2D array. The main method should define the two initial arrays and pass them to the addMatrices method. The method
+should check that the two arrays have the same dimensions, otherwise returning null. To add two arrays you take the
+elements at the same index, add them together, and place the new number at the same index in the new array. For the sake
+of simplicity, assume that the rows in an array have a consistent length. In the main method, if the method returned null,
+print an error message and otherwise print the newly created 2D array.
+ */
```
SamamthaH, line 10: Add a sample output

SamamthaH (changes requested):

src/arrays/multidimensional_arrays/hard/q6/OneAndOneMakeTwoDArrays.java
```diff
@@ -0,0 +1,47 @@
+package arrays.multidimensional_arrays.hard.q6;
+
+/*
+Make a program that defines two 2D integer arrays and uses a method called addMatrices to add them together to create a
+new 2D array. The main method should define the two initial arrays and pass them to the addMatrices method. The method
+should check that the two arrays have the same dimensions, otherwise returning null. To add two arrays you take the
+elements at the same index, add them together, and place the new number at the same index in the new array. For the sake
+of simplicity, assume that the rows in an array have a consistent length. In the main method, if the method returned null,
+print an error message and otherwise print the newly created 2D array.
```
SamamthaH, line 9:
```suggestion
print an error message and otherwise print the newly created 2D array.

```

SamamthaH (approved):

PR #492, opened 2022-12-17
arrays-referencing_array_and_its_elements-hard-q9

KaiKhademi:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/arrays/referencing_array_and_its_elements/hard/q9/Lottery.java
```diff
@@ -0,0 +1,28 @@
+package arrays.referencing_array_and_its_elements.hard.q9;
+
+/*
+Write a Java program that randomly makes an array of integers with size 7 that have 5 digits representing ticket numbers
```
SamamthaH, line 4: Change to:
Write a Java program that randomly generates an array of integers with 7 elements. Each element should have 5 digits representing ticket numbers for a prize draw. Print out the list of tickets then generate a random index for the array and display the index and corresponding ticket number as the winning ticket.

src/arrays/referencing_array_and_its_elements/hard/q9/Lottery.java
```diff
@@ -0,0 +1,28 @@
...
+    12312 24123 13123 45765 45626 13373 13153
+    The winner is the ticket number 3 with number 13123.
+ */
+
+public class Lottery {
+    public static void main(String[] args) {
+        int n = 7;
+        int[] arr = new int[n];
+        for (int i = 0; i < n; i++)
+            arr[i] = 10000 + (int) (Math.random() * (100000 - 10000));
```
SamamthaH, line 19: why the 10000-10000 instead of 90000?

SamamthaH: not that it super matters since its the solution code

src/arrays/referencing_array_and_its_elements/hard/q9/MainTest.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    public void runMain() {
+        Lottery.main(new String[0]);
+    }
+
+    @Test
+    void printsCorrectOutput() {
+        runMain();
+        assertEquals(Integer.parseInt(getItemByName("Ans")), Integer.parseInt(getItemByName("" + (Integer.parseInt(getItemByName("Index"))))),
```
SamamthaH, line 45:
```suggestion
        assertEquals(Integer.parseInt(getItemByName("Ans")), Integer.parseInt(getItemByName("Index"))),
```

src/arrays/referencing_array_and_its_elements/hard/q9/MainTest.java
```diff
@@ -0,0 +1,48 @@
...
+                "Your program does not print the correct ticket number corresponding the selected index.");
```
SamamthaH, line 46:
```suggestion
                "Your program does not print the ticket number from the selected index.");
```

SamamthaH (changes requested):

src/arrays/referencing_array_and_its_elements/hard/q9/Lottery.java
KaiKhademi: It's just easier to understand for someone who reads the code about what it means, in my opinion.

src/arrays/referencing_array_and_its_elements/hard/q9/Lottery.java
```diff
@@ -0,0 +1,28 @@
+package arrays.referencing_array_and_its_elements.hard.q9;
+
+/*
+Write a Java program that randomly generates an array of integers with 7 elements. Each element should have 5 digits
```
SamamthaH, line 4: Because of our discussion in the discord on forcing users to make arrays instead of getting around it just using a few variables we're going to have to kick up the number from 7 to somewhere between 100-500

SamamthaH (changes requested):

src/arrays/referencing_array_and_its_elements/hard/q9/Lottery.java
```diff
@@ -0,0 +1,28 @@
+package arrays.referencing_array_and_its_elements.hard.q9;
+
+/*
+Write a Java program that randomly generates an array of integers with 120 elements. Each element should have 5 digits
+representing ticket numbers for a prize draw. Print out the list of tickets then generate a random index for the array
+and display the index and corresponding ticket number as the winning ticket.
```
SamamthaH, line 6:
```suggestion
and display the index plus 1, and the corresponding ticket number of the winning ticket.
```

SamamthaH (approved):

PR #501, opened 2023-01-26
arrays-programs_involving_multidimensional_data-easy-q5

KaiKhademi:
## PR Checklist
- [ ]  EOF newlines
---
## Website

### DIFF
--- a/src/arrays/programs_involving_multidimensional_data/easy/q5/ValueLock.java
+++ b/src/arrays/programs_involving_multidimensional_data/easy/q5/ValueLock.java
@@ -1,0 +1,22 @@
+package arrays.programs_involving_multidimensional_data.easy.q5;
+
+/*
+Write a Java program that generates a 2D array of size 5 by 5, with random integers in the range of 1 and 999
+(inclusive of both values). Then, print out those values with each row on its own line and the numbers separated by a
+space.
+ */
+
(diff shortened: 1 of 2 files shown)


src/arrays/programs_involving_multidimensional_data/easy/q5/ValueLock.java
```diff
@@ -0,0 +1,22 @@
+package arrays.programs_involving_multidimensional_data.easy.q5;
+
+/*
+Write a Java program that generates a 2D array of size 5 by 5, with random integers in the range of 1 and 999
```
SamamthaH, line 4:
```suggestion
Write a Java program that generates a 5 by 5 2D array, with random integers ranging from 1 to 999
```

src/arrays/programs_involving_multidimensional_data/easy/q5/ValueLock.java
```diff
@@ -0,0 +1,22 @@
...
+(inclusive of both values). Then, print out those values with each row on its own line and the numbers separated by a
```
SamamthaH, line 5:
```suggestion
(inclusive on both ends). Then, print out those values with each row on its own line and the numbers separated by a
```

SamamthaH (approved):

PR #508, opened 2023-01-31
arrays-programs_involving_multidimensional_data-hard-q3

KaiKhademi:
## PR Checklist
- [x]  EOF newlines
---
## Website


Hedgemon4 (changes requested):
See discord messages

src/arrays/programs_involving_multidimensional_data/hard/q3/EightByEight.java
```diff
@@ -0,0 +1,43 @@
+package arrays.programs_involving_multidimensional_data.hard.q3;
+
+/*
+Write a Java program that initializes an 8 by 8 2D array with random values from 0 to 10 (inclusive of both values) and
+prints each row in a line. Then write a function called to find that takes a 2d array
+and an index, and prints out the value at that index. Using that, find the value of the element at index 7.
```
Hedgemon4, line 6: This part of the question text is very confusing

Hedgemon4: Can you fix try and improve this please?

src/arrays/programs_involving_multidimensional_data/hard/q3/EightByEight.java
```diff
@@ -0,0 +1,43 @@
...
+Consider some examples with an 8 by 8 array. Index 2 points to the element in the first row and the third column. Index
+8 points to the element in the second row and the first column. Index 16 points to the element in the third row and the
+first column.
```
Hedgemon4, line 9: We also need to mention that their method should be general enough to work with any input length if we are going to test for that.

src/arrays/programs_involving_multidimensional_data/hard/q3/MainTest.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Test
+    public void printsCorrectOutput() {
+        int x = idx / 8, y = idx % 8;
+        CustomAssertions._assertEquals(Integer.parseInt(getItemByName(x + " " + y)), Integer.parseInt(getItemByName("ans")),
+                "Your program does not correctly print the corresponding element of the index in the 2D array.");
+    }
+
+    static Stream<Arguments> inputProvider() {
```
Hedgemon4, line 58: Can we use arrayUtil here to generate some test cases?

Hedgemon4: And these are the same arrays used in some earlier questions. Please try to tailer the test cases to the question.

Hedgemon4: For example, some edge cases would be the edges of the rows, the first and last items, etc.

Hedgemon4 (changes requested):

src/arrays/programs_involving_multidimensional_data/hard/q3/EightByEight.java
```diff
@@ -0,0 +1,43 @@
...
+Note that your find method should be general enough to work with any size of a 2D array.
+
+Sample Output:
```
Hedgemon4, line 12: Can we add some other information to the output to separate these and make them less like a blob of numbers? Something like "Generated Array: ", and "Value at index x: ".

Hedgemon4 (changes requested):

src/arrays/programs_involving_multidimensional_data/hard/q3/EightByEight.java
```diff
@@ -0,0 +1,48 @@
+package arrays.programs_involving_multidimensional_data.hard.q3;
+
+/*
+Write a Java program that initializes an 8 by 8 2D array with random values from 0 to 10 (inclusive of both values) and 
+generates a random index between 0 and 63 (exclusive), printing both to the console. Then write a function called find
```
Hedgemon4, line 5:
```suggestion
generates a random index between 0 and 63 (exclusive), printing both to the console. Then, write a function called find 
```

Hedgemon4 (approved):

PR #509, opened 2023-01-31
arrays-programs_involving_multidimensional_data-hard-q4

KaiKhademi:
## PR Checklist
- [x]  EOF newlines
---
## Website


Hedgemon4 (changes requested):
See discord messages

src/arrays/programs_involving_multidimensional_data/hard/q4/TicTac.java
```diff
@@ -0,0 +1,28 @@
...
+the shapes to each square. Next, the program prints out the shapes on the game board.
+
+Sample Output:
+    1 0 0
+    1 0 1
+    0 1 0
+ */
+
+public class TicTac {
+    public static void main(String[] args) {
```
Hedgemon4, line 16: We can't guarantee that they actually use an array if we do it this way, so you should add a method to generate this random board instead.

src/arrays/programs_involving_multidimensional_data/hard/q4/TicTac.java
```diff
@@ -0,0 +1,28 @@
+package arrays.programs_involving_multidimensional_data.hard.q4;
+
+/*
+In the game of Tic-Tac-Toe is played on a board of size 3 by 3 between two players. Each player has a shape that can be
+used (either a circle '0' or a cross '1'), and the player has to get three of the same shapes in a row, column, or a
+diagonal to win. Write a Java program that uses a 2D character array to represent the board and randomly assigns one of
+the shapes to each square. Next, the program prints out the shapes on the game board.
+
+Sample Output:
```
Hedgemon4, line 9: In the output, I think we should get them to print an X and O instead of 1 and 0

Hedgemon4 (changes requested):

PR #511, opened 2023-02-01
arrays-programs_involving_multidimensional_data-hard-q6

KaiKhademi:
## PR Checklist
- [x]  EOF newlines
---
## Website


src/arrays/programs_involving_multidimensional_data/hard/q6/MaxVal.java
```diff
@@ -0,0 +1,27 @@
+package arrays.programs_involving_multidimensional_data.hard.q6;
+
+/*
+Write a Java program that initializes a 2D array of random integers of size 5 by 5. All random values generated will be
```
Hedgemon4, line 4: We should add a method component to this question so we can make sure they are actually using an array (probably finding the maximum value).

src/arrays/programs_involving_multidimensional_data/hard/q6/MainTest.java
```diff
@@ -0,0 +1,56 @@
...
+        c[t] = new Optional(new NewLine());
+
+        return c;
+    }
+
+    public void runMain() {
+        MaxVal.main(new String[0]);
+    }
+
+    @Test
```
Hedgemon4, line 43: Just a note, for questions like this, we should run it multiple times

Hedgemon4 (changes requested):

src/arrays/programs_involving_multidimensional_data/hard/q6/MaxVal.java
```diff
@@ -0,0 +1,36 @@
...
+in a range from 0 - 99 (inclusive of both ends). Print out that array in a 5x5 grid, with a space between each number
+and “\n” each row. Then write a method called maxValue that takes a 2D array and return the maximum value of that array.
+Then using the maxValue method print the highest value of the generated 2D array. Your method should be general enough
```
Hedgemon4, line 7:
```suggestion
Then using the maxValue method print the maximum value in the generated 2D array. Your method should be general enough
```

src/arrays/programs_involving_multidimensional_data/hard/q6/MainTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void correctMaxValueMethod(int[][] a, int ans) throws Throwable {
+        Object[][] arguments = {
+                {a, int[][].class},
+        };
+        MethodTest m = new MethodTest(MaxVal.class, "maxValue", arguments);
+        Object output = m.callMethod();
+        CustomAssertions._assertEquals(ans, output, "Your maxValue method does not correctly find the maximum value of a 2D array.");
```
Hedgemon4, line 106:
```suggestion
        CustomAssertions._assertEquals(ans, output, "Your maxValue method does not correctly find the maximum value in a 2D array.");
```

src/arrays/programs_involving_multidimensional_data/hard/q6/MainTest.java
```diff
@@ -0,0 +1,108 @@
...
+                if (max < a)
+                    max = a;
+            }
+        }
+        int out = Integer.parseInt(getItemByName("max"));
+        CustomAssertions._assertEquals(max, out, "Your program does not correctly find the maximum value of the 2D array.");
+    }
+
+
+    static Stream<Arguments> inputProvider() {
```
Hedgemon4, line 67: For these questions, you need to remember to test edge cases. So for example, you want to make sure they are reading all the values in the array, so you would put the max at the very beginning and the very end.

Hedgemon4: Also, you should have some test cases where the 2D array is not square.

Hedgemon4 (changes requested):

src/arrays/programs_involving_multidimensional_data/hard/q6/MaxVal.java
```diff
@@ -0,0 +1,46 @@
+package arrays.programs_involving_multidimensional_data.hard.q6;
+
+/*
+Write a Java program that initializes a 2D array of random integers of size 5 by 5. All random values generated will be
+in a range from 0 - 99 (inclusive of both ends). Print out that array in a 5x5 grid, with a space between each number
+and “\n” each row. Then write a method called maxValue that takes a 2D array and return the maximum value of that array.
```
Hedgemon4, line 6:
```suggestion
and a new line between each row. Then write a method called maxValue that takes a 2D array and returns the maximum value of that array.
```

src/arrays/programs_involving_multidimensional_data/hard/q6/MaxVal.java
```diff
@@ -0,0 +1,46 @@
...
+Then using the maxValue method print the maximum value in the generated 2D array. Your method should be general enough
```
Hedgemon4, line 7:
```suggestion
Using the maxValue method, print the maximum value in the generated 2D array. Note that your method should be general enough
```

src/arrays/programs_involving_multidimensional_data/hard/q6/MainTest.java
```diff
@@ -0,0 +1,133 @@
...
+        int max = -1;
+        for (int i = 0; i < n; i++) {
+            for (int j = 0; j < m; j++) {
+                int a = Integer.parseInt(getItemByName(i + " " + j));
+                if (max < a)
+                    max = a;
+            }
+        }
+        int out = Integer.parseInt(getItemByName("max"));
+        CustomAssertions._assertEquals(max, out, "Your program does not correctly find the maximum value of the 2D array.");
```
Hedgemon4, line 67:
```suggestion
        CustomAssertions._assertEquals(max, out, "Your program does not correctly find and print the maximum value of the 2D array.");
```

src/arrays/programs_involving_multidimensional_data/hard/q6/MainTest.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    static Stream<Arguments> inputProvider() {
+        int[][] t = gen(10, 15);
+        int max = -1;
+        for (int i = 0; i < 10; i++)
+            for (int j = 0; j < 15; j++)
+                max = Math.max(max, t[i][j]);
+        return Stream.of(
+                Arguments.of(
```
Hedgemon4, line 84: You should change some of these test cases so they are not just strictly increasing/decreasing arrays because they test the same thing.

src/arrays/programs_involving_multidimensional_data/hard/q6/MainTest.java
```diff
@@ -0,0 +1,133 @@
...
+                        new int[][]{
+                                {1, 2, 3},
+                                {4, 5, 4},
+                                {3, 2, 1}
+                        },
+                        5
+                ),
+                Arguments.of(
```
Hedgemon4, line 92: Like this case

src/arrays/programs_involving_multidimensional_data/hard/q6/MainTest.java
```diff
@@ -0,0 +1,133 @@
...
+                        new int[][]{
+                                {12, 11, 10, 9},
+                                {8, 7, 6, 5},
+                                {4, 3, 2, 1}
+                        },
+                        12
+                ),
+                Arguments.of(
```
Hedgemon4, line 108: And this case are effectively the same case, so you don't need to test them both.

Hedgemon4: It would be better to have a test case with larger numbers mixed in and put the largest value in a more obscure spot (like not the first, last, or middle spot.

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #512, opened 2023-02-02
arrays-programs_involving_multidimensional_data-hard-q7

KaiKhademi:
## PR Checklist
- [x]  EOF newlines
---
## Website

### DIFF
--- a/src/arrays/programs_involving_multidimensional_data/hard/q7/MainTest.java
+++ b/src/arrays/programs_involving_multidimensional_data/hard/q7/MainTest.java
@@ -1,0 +1,102 @@
+package arrays.programs_involving_multidimensional_data.hard.q7;
+
+import global.BaseRandomTest;
+import global.MethodTest;
+import global.tools.CustomAssertions;
+import global.variables.Clause;
+import global.variables.clauses.*;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
+
+public class MainTest extends BaseRandomTest {
+    // Java
+
+    public static final int n = 5;
+    public static final int up = 100;
+    public static final int down = 0;
+
+    public Clause[] testSentence() {
+        Clause[] c = new Clause[4 * n * n + 2 * n + 4];
+        int t = 0;
+        c[t++] = new StringLiteral("The random 2D array:");
+        c[t++] = new NewLine();
+        for (int i = 0; i < n; i++) {
+            for (int j = 0; j < n; j++) {
+                c[t++] = new RandomDouble(down, up + 1, i + " " + j);
+                c[t++] = new StringLiteral(" ");
+            }
+            c[t++] = new NewLine();
+        }
+        c[t++] = new StringLiteral("The rounded 2D array:");
+        c[t++] = new NewLine();
+        for (int i = 0; i < n; i++) {
+            for (int j = 0; j < n; j++) {
+                c[t++] = new IntegerLiteral(i + "x" + j);
+                c[t++] = new StringLiteral(" ");
+            }
+            c[t++] = new NewLine();
+        }
+        return c;
+    }
+
+    public void runMain() {
+        Rounding.main(new String[0]);
+    }
+
+    @Test
+    public void printsCorrectOutput() {
+        int[][] ans = new int[n][n];
+        int[][] out = new int[n][n];
+        for (int i = 0; i < n; i++) {
+            for (int j = 0; j < n; j++) {
+                ans[i][j] = (int) (Double.parseDouble(getItemByName(i + " " + j)));
+                out[i][j] = Integer.parseInt(getItemByName(i + "x" + j));
+            }
+        }
+        CustomAssertions._assertArrayEquals(ans, out, "Your program does not correctly round the grades in the 2D array.");
+    }
+
+    static Stream<double[][]> inputProvider() {
+        double[][] t1 = {
+                {1.5, 2.4, 3.3, 4.2, 5.1},
+                {2.4, 3.3, 4.2, 5.1, 1.5},
+                {3.3, 4.2, 5.1, 1.5, 2.4},
+                {4.2, 5.1, 1.5, 2.4, 3.3},
+                {5.1, 1.5, 2.4, 3.3, 4.2}
+        };
+        double[][] t2 = {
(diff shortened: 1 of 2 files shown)


src/arrays/programs_involving_multidimensional_data/hard/q7/MainTest.java
```diff
@@ -0,0 +1,102 @@
...
+        for (int i = 0; i < n; i++) {
+            for (int j = 0; j < n; j++) {
+                ans[i][j] = (int) (Double.parseDouble(getItemByName(i + " " + j)));
+                out[i][j] = Integer.parseInt(getItemByName(i + "x" + j));
+            }
+        }
+        CustomAssertions._assertArrayEquals(ans, out, "Your program does not correctly round the grades in the 2D array.");
+    }
+
+    static Stream<double[][]> inputProvider() {
```
Hedgemon4, line 62: If we test cases where the array length is different than the question, we need to let the student know that their method should be general enough to work with any array length so they don't hard code it for length five or something.

src/arrays/programs_involving_multidimensional_data/hard/q7/Rounding.java
```diff
@@ -0,0 +1,55 @@
+package arrays.programs_involving_multidimensional_data.hard.q7;
+
+/*
+Suppose you have a 2D array that stores student assignment grades. In particular, each row represents the grades for a
```
Hedgemon4, line 4: Also, let's use Math.round here instead of truncating the number.

src/arrays/programs_involving_multidimensional_data/hard/q7/MainTest.java
```diff
@@ -0,0 +1,102 @@
...
+    public void printsCorrectOutput() {
+        int[][] ans = new int[n][n];
+        int[][] out = new int[n][n];
+        for (int i = 0; i < n; i++) {
+            for (int j = 0; j < n; j++) {
+                ans[i][j] = (int) (Double.parseDouble(getItemByName(i + " " + j)));
+                out[i][j] = Integer.parseInt(getItemByName(i + "x" + j));
+            }
+        }
+        CustomAssertions._assertArrayEquals(ans, out, "Your program does not correctly round the grades in the 2D array.");
```
Hedgemon4, line 59: Should make the error message include something about printing as well, like "... round and print the grades ... "

src/arrays/programs_involving_multidimensional_data/hard/q7/MainTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
+
+public class MainTest extends BaseRandomTest {
+    // Java
+
+    public static final int n = 5;
+    public static final int up = 100;
```
Hedgemon4, line 18: Can you rename these to upperBound and lowerBound please? I think the names are too unclear currently

src/arrays/programs_involving_multidimensional_data/hard/q7/MainTest.java
```diff
@@ -0,0 +1,102 @@
...
+        };
+
+        return Stream.of(
+                t1, t2, t3, t4
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    public void correctRoundingMethod(double[][] arr) throws Throwable {
```
Hedgemon4, line 89: You should have some hardcoded cases with the founded array instead of the helper method here

Hedgemon4: Helper methods are mainly for when we provide generated input

src/arrays/programs_involving_multidimensional_data/hard/q7/Rounding.java
```diff
@@ -0,0 +1,55 @@
...
+different assignment from all the students. You have to enter the student grades into a database and suddenly realize
+that the database only takes integers and not decimal numbers. So you will round the grade down for each student. Write
+a program that carries this conversion out. Write a Java Program , that makes a 5 * 5 2D array and fills it with random
```
Hedgemon4, line 7:
```suggestion
Write a Java Program that makes a 5 by 5 2D array, fills it with random
```

Hedgemon4: This is a comma splice

src/arrays/programs_involving_multidimensional_data/hard/q7/Rounding.java
Hedgemon4, line 4: You can probably remove the second sentence from the question

src/arrays/programs_involving_multidimensional_data/hard/q7/Rounding.java
```diff
@@ -0,0 +1,55 @@
+package arrays.programs_involving_multidimensional_data.hard.q7;
+
+/*
+Suppose you have a 2D array that stores student assignment grades. In particular, each row represents the grades for a
+different assignment from all the students. You have to enter the student grades into a database and suddenly realize
+that the database only takes integers and not decimal numbers. So you will round the grade down for each student. Write
```
Hedgemon4, line 6:
```suggestion
that the database only takes integers, so you will need to round the student's grades.
```

src/arrays/programs_involving_multidimensional_data/hard/q7/Rounding.java
```diff
@@ -0,0 +1,55 @@
...
+a program that carries this conversion out. Write a Java Program , that makes a 5 * 5 2D array and fills it with random
+numbers between 0 to 100, and prints it. Then create a called rounding function to round down each of the grades in the
```
Hedgemon4, line 8:
```suggestion
numbers between 0 to 100, and prints it. Then create a called rounding function to round each of the grades in the
```

src/arrays/programs_involving_multidimensional_data/hard/q7/Rounding.java
```diff
@@ -0,0 +1,55 @@
...
+array to the floor, then return that integer array. Finally, print the rounded 2D array.
```
Hedgemon4, line 9:
```suggestion
array, and returns that integer array. Finally, print the rounded 2D array.
```

Hedgemon4 (changes requested):

PR #515, opened 2023-02-05
pre_defined_classes-programs_with_changing_behaviors-hard-q11

te-sa:
## PR Checklist
- [x]  EOF newlines
---
## Website

### DIFF
--- a/src/pre_defined_classes/programs_with_changing_behaviour/medium/q11/MainTest.java
+++ b/src/pre_defined_classes/programs_with_changing_behaviour/medium/q11/MainTest.java
@@ -8,9 +8,7 @@ import org.junit.jupiter.api.RepeatedTest;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 public class MainTest extends BaseRandomTest {
-
-    @Override
-    // define what output should look like
+    // Java
     public Clause[] testSentence() {
         return new Clause[]{
                 new StringLiteral("Your score for this round was "),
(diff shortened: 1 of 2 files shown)


src/pre_defined_classes/programs_with_changing_behaviour/medium/q11/MainTest.java
```diff
@@ -0,0 +1,42 @@
...
+import global.BaseRandomTest;
+import global.variables.Clause;
+import global.variables.clauses.*;
+import org.junit.jupiter.api.RepeatedTest;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MainTest extends BaseRandomTest {
+
+    @Override
```
Hedgemon4, line 12: We usually get rid of these overrides by convention

src/pre_defined_classes/programs_with_changing_behaviour/medium/q11/MainTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+import global.BaseRandomTest;
+import global.variables.Clause;
+import global.variables.clauses.*;
+import org.junit.jupiter.api.RepeatedTest;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MainTest extends BaseRandomTest {
+
```
Hedgemon4, line 11: We also add a comment at the start to indicate what kind of question this is. If it is a java, we put `// Java`, if it is a Parsons question we put `// Parsons`, and if it is a Parsons with distractors we put `// Parsons with distractors`

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #516, opened 2023-02-05
conditionals-programs_with_decision_points-hard-q12

te-sa:
## PR Checklist
- [ ]  EOF newlines

### DIFF
--- a/src/conditionals/programs_with_decision_points/hard/q12/MainTest.java
+++ b/src/conditionals/programs_with_decision_points/hard/q12/MainTest.java
@@ -17,7 +17,6 @@ import java.util.stream.Stream;
 public class MainTest extends BaseTest {
     // Java
 
-    @Override
     public Clause[] testSentence() {
         // when you need to provide Scanner input
         TestOption.isInputTest = true;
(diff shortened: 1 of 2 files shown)


src/conditionals/programs_with_decision_points/hard/q12/MainTest.java
```diff
@@ -0,0 +1,94 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
+
+public class MainTest extends BaseTest {
+    // Java
+
+    @Override
```
Hedgemon4, line 20: Same thing here, you can remove the Override

src/conditionals/programs_with_decision_points/hard/q12/MainTest.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    public void runMain() {
+        SignFromTheStars.main(new String[0]);
+    }
+
+    @ParameterizedTest
+    @MethodSource("InputProvider")
+    void printsAstrologicalSignCorrectly(int day, int month, String sign) throws InvalidClauseException {
```
Hedgemon4, line 88: I would name this `printsCorrectAstrologicalSign`, I feel like it conveys that it prints the correct option better

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #517, opened 2023-02-06
methods-modular-programs-hard-q10

te-sa:
## PR Checklist
- [x]  EOF newlines

### DIFF
--- a/src/methods/modular_programs/hard/q10/MainTest.java
+++ b/src/methods/modular_programs/hard/q10/MainTest.java
@@ -40,16 +40,28 @@ public class MainTest extends BaseTest {
     static Stream<Arguments> inputProvider() {
         return Stream.of(
                 Arguments.of(1960, "Rat"),
+                Arguments.of(1972, "Rat"),
                 Arguments.of(1985, "Ox"),
+                Arguments.of(2033, "Ox"),
                 Arguments.of(1962, "Tiger"),
+                Arguments.of(1950, "Tiger"),
                 Arguments.of(1951, "Rabbit"),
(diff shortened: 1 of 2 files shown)


src/methods/modular_programs/hard/q10/MainTest.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    public void runMain() {
+        GoatYear.main(new String[0]);
+    }
+
+    static Stream<Arguments> inputProvider() {
+        return Stream.of(
+                Arguments.of(1960, "Rat"),
```
Hedgemon4, line 42: I would say that you should put two cases for each year

Hedgemon4 (changes requested):

src/methods/modular_programs/hard/q10/MainTest.java
```diff
@@ -0,0 +1,77 @@
...
+                Arguments.of(1994, "Dog"),
+                Arguments.of(1959, "Pig"),
+                Arguments.of(603, "Pig")
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void printsCorrectChineseZodiacSign(int year, String sign) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not print the correct chinese zodiac sign for the entered year.";
```
Hedgemon4, line 71:
```suggestion
        TestOption.incorrectStructureErrorMessage = "Your program does not print the correct Chinese zodiac sign for the entered year.";
```

Hedgemon4 (approved):

PR #518, opened 2023-02-06
arrays-programs_involving_data_sequences-hard-q9

te-sa:
## PR Checklist
- [ ]  EOF newlines

### DIFF
--- a/src/arrays/programs_involving_data_sequences/hard/q9/MainTest.java
+++ b/src/arrays/programs_involving_data_sequences/hard/q9/MainTest.java
@@ -1,0 +1,79 @@
+package arrays.programs_involving_data_sequences.hard.q9;
+
+import global.BaseTest;
+import global.variables.Clause;
+import global.variables.clauses.IntegerLiteral;
+import global.variables.clauses.NewLine;
+import global.variables.clauses.PlaceHolder;
+import global.variables.clauses.StringLiteral;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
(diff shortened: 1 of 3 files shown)


src/arrays/programs_involving_data_sequences/hard/q9/MainTest.java
```diff
@@ -0,0 +1,79 @@
...
+            "Nunavut",
+            "Ontario",
+            "Prince Edward Island",
+            "Quebec",
+            "Saskatchewan",
+            "Yukon"
+    };
+
+    @Override
+    public Clause[] testSentence() {
```
Hedgemon4, line 37: The problem with this PR is that you didn't declare it was an input test, that is why it is failing

Hedgemon4 (changes requested):

src/arrays/programs_involving_data_sequences/hard/q9/MainTest.java
te-sa: I made that change and it is still failing

src/arrays/programs_involving_data_sequences/hard/q9/QuizGameEh.java
```diff
@@ -0,0 +1,88 @@
...
+                "Toronto",
+                "Charlottetown",
+                "Quebec City",
+                "Regina",
+                "Whitehorse"
+        };
+        quiz(jurisdictions, answers);
+    }
+
+    public static void quiz(String[] jurisdictions, String[] answers) {
```
Hedgemon4, line 76: You need to add a method test for this

Hedgemon4 (changes requested):

PR #519, opened 2023-02-06
arrays-programs_involving_data_sequences-hard-q10

te-sa:
## PR Checklist
- [x]  EOF newlines

### DIFF
--- a/src/arrays/programs_involving_data_sequences/hard/q10/ArrayThereDelilah.java
+++ b/src/arrays/programs_involving_data_sequences/hard/q10/ArrayThereDelilah.java
@@ -5,7 +5,7 @@ Create a Java program that takes an array of strings, finds the middle character
 console. To do this, create a method called middle, which accepts a String array, finds the middle character in the
 string, and adds it to a character array. For this method, find the middle element by dividing the length by two,
 rounding the number, and subtracting one. After computing the array, return it and print each character in the main
-method. Use the provided main method to call your method and print the output.
+method. Use the provided main method to call your method and print the output. Note that your middle method should be general enough to work with input of any length.
 
 Sample Output:
         The middle letter of the array are:
--- a/src/arrays/programs_involving_data_sequences/hard/q10/ArrayThereDelilah.java
+++ b/src/arrays/programs_involving_data_sequences/hard/q10/ArrayThereDelilah.java
@@ -30,5 +30,4 @@ public class ArrayThereDelilah {
             middle[i] = arr[i].charAt((int) Math.round(arr[i].length() / 2.0) - 1);
         return middle;
     }
-
 }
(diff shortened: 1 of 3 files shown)


src/arrays/programs_involving_data_sequences/hard/q10/MainTest.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    public void runMain() {
+        ArrayThereDelilah.main(new String[0]);
+    }
+
+    static Stream<Arguments> inputProvider() {
+        return Stream.of(
+                Arguments.of(new String[]{"add", "here", "lilly", "president", "jelly", "advance", "archers"}, "delilah".toCharArray()),
+                Arguments.of(new String[]{"archers", "get", "hello", "island", "cloud"}, "hello".toCharArray()),
+                Arguments.of(new String[]{"fireworks", "acorn", "array", "pilot", "hedge"}, "world".toCharArray()),
```
Hedgemon4, line 36: You can also take a look at arrayUtil for this question. There is a method that will generate an array of random words for you, which would be useful for this method. However, if you did that you would have to create a helper method, but that should be fine as long as you don't look at the original code.

Hedgemon4 (changes requested):

src/arrays/programs_involving_data_sequences/hard/q10/MainTest.java
te-sa: I made that change but also kept the hardcoded ones.

src/arrays/programs_involving_data_sequences/hard/q10/MainTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void correctMiddleMethod(String[] input, char[] word) throws Throwable {
+        Object[][] arguments = {
+                {input, String[].class},
+        };
+        MethodTest m = new MethodTest(ArrayThereDelilah.class, "middle", arguments);
+        Object output = m.callMethod();
+        CustomAssertions._assertArrayEquals(word, output, "Your middle method does not return the correct array of middle characters.");
```
Hedgemon4, line 67:
```suggestion
        CustomAssertions._assertArrayEquals(word, output, "Your middle method does not a character array with the middle character of each word.");
```

src/arrays/programs_involving_data_sequences/hard/q10/ArrayThereDelilah.java
```diff
@@ -0,0 +1,33 @@
+package arrays.programs_involving_data_sequences.hard.q10;
+
+/*
+Create a Java program that takes an array of strings, finds the middle character in each string, and prints it to the
+console. To do this, create a method called middle, which accepts a String array, finds the middle character in the
+string, and adds it to a character array. For this method, find the middle element by dividing the length by two,
+rounding the number, and subtracting one. After computing the array, return it and print each character in the main
+method. Use the provided main method to call your method and print the output.
```
Hedgemon4, line 8:
```suggestion
method. Use the provided main method to call your method and print the output. Note that your middle method should be general enough to work with input of any length.
```

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #524, opened 2023-02-13
oop-programs_with_multiple_classes-medium-q1

Hedgemon4:
## PR Checklist
- [ ]  EOF newlines

### DIFF
--- a/src/oop/programs_with_multiple_classes/medium/q1/MainTest.java
+++ b/src/oop/programs_with_multiple_classes/medium/q1/MainTest.java
@@ -1,5 +1,44 @@
 package oop.programs_with_multiple_classes.medium.q1;
 
+import global.ObjectTest;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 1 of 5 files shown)


src/oop/programs_with_multiple_classes/medium/q1/MainTest.java
```diff
@@ -0,0 +1,201 @@
...
+        String tableClassString = "oop.programs_with_multiple_classes.medium.q1.Table";
+        String testRoomClassString = "oop.programs_with_multiple_classes.medium.q1.TestRoom";
+        chair = new ObjectTest(chairClassString);
+        table = new ObjectTest(tableClassString);
+        testRoom = new ObjectTest(testRoomClassString);
+    }
+
+    public Clause[] testSentence() {
+        TestOption.incorrectStructureErrorMessage =
+                "Your program does not correctly print out the toString for each initialized object in the TestRoom class";
```
Hedgemon4, line 39:
```suggestion
                "Your program does not print out the correct string for each initialized object in the TestRoom class.";
```

PR #533, opened 2023-02-13
fix-arrays-programs_involving_data_sequences-easy-q2

KaiKhademi:
## PR Checklist
- [ ]  EOF newlines

### DIFF
--- a/src/arrays/programs_involving_data_sequences/easy/q2/MainTest.java
+++ b/src/arrays/programs_involving_data_sequences/easy/q2/MainTest.java
@@ -1,11 +1,14 @@
 package arrays.programs_involving_data_sequences.easy.q2;
 
 import global.BaseTest;
+import global.MethodTest;
+import global.tools.TestOption;
 import global.utils.ArrayUtil;
 import global.utils.MethodUtil;
 import global.variables.Clause;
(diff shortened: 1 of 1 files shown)


src/arrays/programs_involving_data_sequences/easy/q2/MainTest.java
```diff
@@ -50,18 +53,25 @@ static Stream<int[]> inputProvider() {
     @ParameterizedTest
     @MethodSource("inputProvider")
     void correctPrintArrayPatternMethod(int[] input) throws Throwable {
-        MethodUtil.invokeIfMethodExists(StarryNight.class, "printArrayPattern", new Object[]{input}, int[].class);
-        String output = MethodUtil.getMethodOutput();
-        assertEquals(pattern(input), output, "Your printArrayPattern method does not correctly print the pattern based on the input array.");
-    }
+        Clause[] c = new Clause[2 * input.length];
```
Hedgemon4, line 56: You should use the `createClauseArray` helper here instead of this code

Hedgemon4 (changes requested):

Hedgemon4 (approved):

Hedgemon4 (approved):

PR #585, opened 2023-03-07
Changed to Java question arrays-multidimensional_arrays-hard-q9

KaiKhademi:
## PR Checklist
- [ ]  EOF newlines


Hedgemon4 (changes requested):
Please resolve the merge conflicts

src/arrays/multidimensional_arrays/hard/q9/ImAWholeDimensionAhead.java
```diff
@@ -1,16 +1,25 @@
 package arrays.multidimensional_arrays.hard.q9;
 
 /*
-Make a java program that defines a 3D array in the main method and then passes it to a method called
-imAWholeDimensionAhead. This method should accept a 3D array and print out each element on the same line.
+Make a Java program that randomly initialize a 3 * 3 * 3 3D array with integers ranging from 1 to 100
```
Hedgemon4, line 4: Same for this question. It will need to be changed as arrays with methods is after this category, so we can't have a method which accepts and processes an array.

Hedgemon4 (changes requested):

SamamthaH:
this is getting moved to arrays with methods so the question itself does not need to be changed

PR #588, opened 2023-03-11
oop-programs_with_multiple_classes-hard-q2

te-sa:
## PR Checklist
- [ ]  EOF newlines


src/oop/programs_with_multiple_classes/hard/q2/QuestionText.md
```diff
@@ -0,0 +1,39 @@
...
+method has been provided in the `Person` and `Blanket` classes for you to use.
+
+## Person Class
+
+Your `Person` class should have the following items:
+
+- Attributes:
+    - A `double` called **height**
+- Constructors:
+    - A `public Person(double)` constructor which initializes the **height** field with the passed double
```
Hedgemon4, line 14: For this, I think Bowen wants us to present constructors and methods differently. Might want to say something more like "A constructor which accepts a double to initialize the height field and is accessible from other classes. "

Hedgemon4: Instead of just outright giving them the method header.

src/oop/programs_with_multiple_classes/hard/q2/QuestionText.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Blanket Class
+
+Your `Blanket` class should have the following items:
+
+- Attributes:
+    - A `String` called **size**
+    - An `int` called **threadCount**
+- Constructors:
+    - A `public Blanket(String, int)` constructor which initializes the **size** field with the passed String and the
```
Hedgemon4, line 24: Same as the other constructor comment

src/oop/programs_with_multiple_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,191 @@
...
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
+
+import static global.tools.CustomAssertions._assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class MainTest extends BaseTest {
+
```
Hedgemon4, line 22: Should also still say what kind of question this is (java, parsons, etc.)

src/oop/programs_with_multiple_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,191 @@
...
+    private ObjectTest person;
+    private ObjectTest blanket;
+
+    private final String height = "height";
```
Hedgemon4, line 26: I think for these fields which describe the names of class fields, instead of naming them by whatever the original name was, we should name them like "personStringFieldName"

src/oop/programs_with_multiple_classes/hard/q2/TestFort.java
```diff
@@ -0,0 +1,13 @@
+package oop.programs_with_multiple_classes.hard.q2;
+
+public class TestFort {
+    public static void main(String[] args) {
```
Hedgemon4, line 4: Can we make this a method instead of the main method? Like, have three methods which each takes the values to pass to the constructor and then return the constructed object?

src/oop/programs_with_multiple_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,191 @@
...
+        TestFort.main(new String[0]);
+    }
+
+    /**
+     * Person Tests
+     **/
+    @Test
+    public void personClassHasCorrectFields() {
+        String missingFieldMessage = "Your " + personLc + " class is missing a required field.";
+        String incorrectFieldModifierMessage = "A field in your " + personLc + " class has an incorrect modifier.";
```
Hedgemon4, line 77: I think we should say an "incorrect visibility modifier"

src/oop/programs_with_multiple_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,191 @@
...
+                {height, double.class}
+        };
+        Object personInstance = person.createInstance(arguments);
+        _assertEquals(height, person.getFieldValue(personInstance, this.height),
+                "Your " + personLc + " constructor does not correctly initialize the " + this.height + " field.");
+    }
+
+    @ParameterizedTest
+    @MethodSource("personInputProvider")
+    public void correctPersonToStringMethod(Double height) throws Throwable {
```
Hedgemon4, line 116: I think we should remove this because ideally we can have that thing were we can configure if they can edit part of a class

Hedgemon4, line 116: Ideally, if we get the thing where we can prevent the students from editing certain portions of the code, we can remove this (but just keep it for now)

src/oop/programs_with_multiple_classes/hard/q2/QuestionText.md
```diff
@@ -0,0 +1,39 @@
...
+You want to write a program that models a group of people and their blanket fort. To do this, create a Java program with
+three classes: `Person`, `Blanket`, and `TestFort`. See the requirements below for each class. Note that a `toString`
+method has been provided in the `Person` and `Blanket` classes for you to use.
+
+## Person Class
+
+Your `Person` class should have the following items:
+
+- Attributes:
+    - A `double` called **height**
```
Hedgemon4, line 12: We should maybe also say something about it not being accessible from another class

Hedgemon4 (changes requested):

src/oop/programs_with_multiple_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,130 @@
...
+        blanket = new ObjectTest(blanketClassString);
+    }
+
+    /**
+     * Person Tests
+     **/
+    @Test
+    public void personClassHasCorrectFields() {
+        String missingFieldMessage = "Your " + personLc + " class is missing a required field.";
+        String incorrectFieldModifierMessage = "A field in your " + personLc + " class has an incorrect modifier.";
```
Hedgemon4, line 44:
```suggestion
        String incorrectFieldModifierMessage = "A field in your " + personLc + " class has the wrong visibility modifier.";
```

src/oop/programs_with_multiple_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,130 @@
...
+        assertTrue(person.hasField(height, double.class), missingFieldMessage);
+        assertTrue(person.hasModifier(height, "private"), incorrectFieldModifierMessage);
+    }
+
+    @Test
+    public void personClassHasCorrectConstructors() {
+        String missingConstructorMessage = "Your " + personLc + " class does not have a required constructor.";
+        String incorrectConstructorModifierMessage =
+                "One of the constructors in your " + personLc + " class does not have the correct modifiers.";
```
Hedgemon4, line 53: Same as above, I don't know what the message is we should use

src/oop/programs_with_multiple_classes/hard/q2/MainTest.java
Hedgemon4: What do you think of this?

src/oop/programs_with_multiple_classes/hard/q2/QuestionText.md
Hedgemon4: You did not address this comment

Hedgemon4: They need to know if the attribute should be public vs private

Hedgemon4 (changes requested):

src/oop/programs_with_multiple_classes/hard/q2/MainTest.java
te-sa: I think that sounds good, good to be specific since we're only checking visibility modifiers here anyways

src/oop/programs_with_multiple_classes/hard/q2/MainTest.java
te-sa: Should I remove this or not? Not sure from the other comment

te-sa: I kept both in for now

src/oop/programs_with_multiple_classes/hard/q2/MainTest.java
te-sa: I put: "One of the constructors in your " + personLc + " class does not have the correct visibility modifier."

src/oop/programs_with_multiple_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,130 @@
...
+        String blanketClassString = "oop.programs_with_multiple_classes.hard.q2." + blanketUc;
+        person = new ObjectTest(personClassString);
+        blanket = new ObjectTest(blanketClassString);
+    }
+
+    /**
+     * Person Tests
+     **/
+    @Test
+    public void personClassHasCorrectFields() {
```
Hedgemon4, line 42:
```suggestion
    public void personClassHasRequiredAttributes() {
```

src/oop/programs_with_multiple_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,130 @@
...
+        _assertEquals(height, person.getFieldValue(personInstance, this.height),
+                "Your " + personLc + " constructor does not correctly initialize the " + this.height + " field.");
+    }
+
+    /**
+     * Blanket Tests
+     **/
+    @Test
+    public void blanketClassHasCorrectFields() {
+        String missingFieldMessage = "Your " + blanketLc + " class is missing a required field.";
```
Hedgemon4, line 86: Please apply the same changes I mentioned in the comments above to your blanket tests.

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #591, opened 2023-03-12
oop-programs_with_multiple_classes-hard-q4

te-sa:
## PR Checklist
- [ ]  EOF newlines


Hedgemon4 (changes requested):
See requested changes on https://github.com/canvas-gamification/canvas-gamification-junit-tests/pull/588

src/oop/programs_with_multiple_classes/hard/q4/QuestionText.md
```diff
@@ -0,0 +1,34 @@
...
+    - An `int` called **year**
+    - A `String` called **type**
+- Constructors:
+    - A constructor which accepts a String to initialize the **type** field and an int to initialize the **year**
+      field (in this order)
+        - This constructor should be accessible from other classes
+- Methods:
+    - A method called `determineTime`
+        - It should take no input and return an int
+        - Its purpose is to calculate the amount of years since the **year** when a book was
```
Hedgemon4, line 20: This definitely needs a hint for the Year.now() part because I did not know that was even a thing

src/oop/programs_with_multiple_classes/hard/q4/Bookcase.java
```diff
@@ -0,0 +1,11 @@
+package oop.programs_with_multiple_classes.hard.q4;
+
+public class Bookcase {
+    private int sizeCapacity;
+    private Book b1;
```
Hedgemon4, line 5: I have to double-check if we can have this kind of object relation in this category. I will ask Bowen and get back to you, but just leave it the way it is for now.

src/oop/programs_with_multiple_classes/hard/q4/QuestionText.md
```diff
@@ -0,0 +1,34 @@
...
+## Book Class
+
+Your `Book` class should have the following items:
+
+- Attributes:
+    - An `int` called **year**
+    - A `String` called **type**
+- Constructors:
+    - A constructor which accepts a String to initialize the **type** field and an int to initialize the **year**
+      field (in this order)
```
Hedgemon4, line 15: Can you format this md file, please? You should be able to use ctrl - alt - l on it like any other file in Intellij

src/oop/programs_with_multiple_classes/hard/q4/MainTest.java
```diff
@@ -0,0 +1,150 @@
...
+        book = new ObjectTest(bookClassString);
+        bookcase = new ObjectTest(bookcaseClassString);
+    }
+
+    /**
+     * Book tests
+     **/
+    @Test
+    public void bookClassHasCorrectFields() {
+        String missingFieldMessage = "Your " + bookLc + " class is missing a required field.";
```
Hedgemon4, line 47: Do we think that the ` is missing a required field` is detailed enough? I did it because I did not want to give the answer away, but it might be too vague for the student.

src/oop/programs_with_multiple_classes/hard/q4/MainTest.java
```diff
@@ -0,0 +1,150 @@
...
+    @MethodSource("bookInputProvider")
+    public void correctDetermineTimeMethod(int year, String type) throws Throwable {
+        Object[][] arguments = new Object[][]{
+                {year, int.class},
+                {type, String.class}
+        };
+        Object bookInstance = book.createInstance(arguments);
+        Object determineTimeOutput = book.callMethod(determineTime, bookInstance);
+        int determineTimeExpected = Year.now().getValue() - year;
+        String incorrectDetermineTimeExpected = "Your " + bookLc + " " + determineTime + " method does not return the correct time.";
```
Hedgemon4, line 101: Don't they have to calculate a year here, not a time?

Hedgemon4 (changes requested):

src/oop/programs_with_multiple_classes/hard/q4/QuestionText.md
te-sa: The idea is that students wouldn't need to know this, they could just use "2023" or whatever the current year is. This will only be used in the test code, and it is in my solution code so the tests will continue to pass in coming years

src/oop/programs_with_multiple_classes/hard/q4/Bookcase.java
te-sa: Did she get back to you?

src/oop/programs_with_multiple_classes/hard/q4/MainTest.java
te-sa: I think it should be. They should know what fields are.

te-sa: We may want to call it either "attribute" or "field" though, currently we're using both. "Attribute" in the question text and "field" in the error message. Which do you prefer?

Hedgemon4: attribute

src/oop/programs_with_multiple_classes/hard/q4/QuestionText.md
Hedgemon4: Ah ok

Hedgemon4: Maybe we can specify to use the current year though?

src/oop/programs_with_multiple_classes/hard/q4/QuestionText.md
```diff
@@ -0,0 +1,34 @@
...
+## Bookcase Class
+
+Your `Bookcase` class should have the following items:
+
+- Attributes:
+    - A `int` called **sizeCapacity**, which is not accessible from other classes
+    - A `Book` called **b1**, which is not accessible from other classes
+- Constructors:
+    - A constructor which accepts an int to initialize the **sizeCapacity** field and a Book to initialize the **b1**
+      field (in this order)
```
Hedgemon4, line 33:
```suggestion
    - A constructor which accepts an int to initialize the **sizeCapacity** attribute and a Book to initialize the **b1**
      attribute (in this order)
```

src/oop/programs_with_multiple_classes/hard/q4/MainTest.java
```diff
@@ -0,0 +1,150 @@
...
+        book = new ObjectTest(bookClassString);
+        bookcase = new ObjectTest(bookcaseClassString);
+    }
+
+    /**
+     * Book tests
+     **/
+    @Test
+    public void bookClassHasCorrectFields() {
+        String missingFieldMessage = "Your " + bookLc + " class is missing a required field.";
```
Hedgemon4, line 47: Bowen asked that we name the missing attribute, and we have been using something similar to the message below:
"The %s class is missing the %s attribute. Make sure that the class contains the field and that it is spelled correctly."

src/oop/programs_with_multiple_classes/hard/q4/MainTest.java
```diff
@@ -0,0 +1,150 @@
...
+        String incorrectFieldModifierMessage = "A field in your " + bookLc + " class has an the wrong visibility modifier.";
+        assertTrue(book.hasField(year, int.class), missingFieldMessage);
```
Hedgemon4, line 49: You can also use the new method I added which allows you to check if the attribute exists before checking if it has the correct type. This way we can give two error messages to the student. One that mentions the attribute could not be found and to check it was added and spelt correctly and another to check that the attribute has the correct type.

src/oop/programs_with_multiple_classes/hard/q4/MainTest.java
```diff
@@ -0,0 +1,150 @@
...
+    @MethodSource("bookInputProvider")
+    public void correctHowManyYearsMethod(int year, String type) throws Throwable {
+        Object[][] arguments = new Object[][]{
+                {year, int.class},
+                {type, String.class}
+        };
+        Object bookInstance = book.createInstance(arguments);
+        Object howManyYearsOutput = book.callMethod(howManyYears, bookInstance);
+        int howManyYearsExpected = Year.now().getValue() - year;
+        String incorrectHowManyYearsExpected = "Your " + bookLc + " " + howManyYears + " method does not return the correct time.";
```
Hedgemon4, line 101: I think it could be more helpful to the student if we said "... the correct number of years." instead of "... correct time". What do you think?

src/oop/programs_with_multiple_classes/hard/q4/MainTest.java
```diff
@@ -0,0 +1,150 @@
...
+        _assertEquals(year, book.getFieldValue(bookInstance, this.year),
+                "Your " + book + " constructor does not correctly initialize the " + this.year + " field.");
+        _assertEquals(type, book.getFieldValue(bookInstance, this.type),
+                "Your " + book + " constructor does not correctly initialize the " + this.type + " field.");
+    }
+
+    @ParameterizedTest
+    @MethodSource("bookInputProvider")
+    public void correctHowManyYearsMethod(int year, String type) throws Throwable {
+        Object[][] arguments = new Object[][]{
```
Hedgemon4, line 94: Also, for methods we decided to add a check if they exist before calling them. That way we can check things like the visibility modifier separately from the actual method definition, and we can also provide a custom error message based on which part is wrong instead of the default message that it could not be found.

src/oop/programs_with_multiple_classes/hard/q4/MainTest.java
```diff
@@ -0,0 +1,150 @@
...
+        Class<?>[] constructorClasses = new Class[]{
+                int.class,
+                Book.class
+        };
+        assertTrue(bookcase.hasConstructor(constructorClasses), missingConstructorMessage);
+        assertTrue(bookcase.hasModifier(constructorClasses, "public"), incorrectConstructorModifierMessage);
+    }
+
+    private static Stream<Arguments> bookcaseInputProvider() {
+        return Stream.of(Arguments.of(1, new Book(2004, "Novella")),
```
Hedgemon4, line 130: I don't think we should do this as it would crash our code if the book constructor does not exist.

Hedgemon4: Instead, you could pass the arguments you want for the book, put the `createInstance` call in a try-catch, and then you should be able to provide a custom error message related to the question that the book object could not be created to pass to the bookcase constructor.

Hedgemon4 (changes requested):

Hedgemon4 (changes requested):
See comment on q3 in this category.

PR #602, opened 2023-03-17
 oop-creating_objects_from_classes-hard-q4

te-sa:
## PR Checklist
- [ ]  EOF newlines


Hedgemon4 (changes requested):
See requested changes on https://github.com/canvas-gamification/canvas-gamification-junit-tests/pull/588

Hedgemon4 (changes requested):
Please review the changes I suggested on the other PRs in this category and apply them here.

src/oop/creating_objects_from_classes/hard/q4/QuestionText.md
```diff
@@ -0,0 +1,44 @@
...
+
+Then, call the `getTotal` method and print the balance of the created Savings object on its own line. The result from 
+doing this should be the printing of the `String` "Balance: x" where x is the value of the total.
+
+Then, call the `recieveBonus` method and repeat the print statement to view the new balance.
+
+Note: You do not need to place the savings class in your submission code, the website will handle this for you. It is 
+just for you to reference.
+
+### Dog Class:
```
Hedgemon4, line 24: I don't think this is the correct class here.

src/oop/creating_objects_from_classes/hard/q4/QuestionText.md
```diff
@@ -0,0 +1,44 @@
...
+
+Suppose you have defined a `Savings` class to model your investment account, and now you want to test it to make sure 
+that it works. Given the savings class below, create a test class called `BankAccount` with a main method and 
+initialize a `Savings` object with the following attribute:
+- Initial: 500.32
+
+Then, call the `getTotal` method and print the balance of the created Savings object on its own line. The result from 
+doing this should be the printing of the `String` "Balance: x" where x is the value of the total.
+
+Then, call the `recieveBonus` method and repeat the print statement to view the new balance.
```
Hedgemon4, line 19: I think "repeate print the statement" is a bit vague. Maybe something more along the lines of " then print the updated balance with the same format as before." or something similar.

src/oop/creating_objects_from_classes/hard/q4/MainTest.java
```diff
@@ -0,0 +1,93 @@
...
+            m.callMethod();
+            double first = Double.parseDouble(m.getMethodItemByName("num1"));
+            firstRandomValues.add(first - initial);
+            secondRandomValues.add(Double.parseDouble(m.getMethodItemByName("num2")) - first);
+        }
+
+        RandomDouble randomDouble = new RandomDouble(bonusMin, bonusMax);
+        assertTrue(randomDouble.validateRandom(firstRandomValues), "Your first balance is not being correctly generated. Please ensure you are setting the correct initial value.");
+        assertTrue(randomDouble.validateRandom(secondRandomValues), "Your second balance is not being correctly generated. Please ensure you have called the " + method2Name + " method.");
+    }
```
Hedgemon4, line 90: I don't think that you need to check the randomness here. The student is not able to tweak the code for the class, so I think you can just record the value that our code generates, and make sure that is the value the test prints. This is more work than we need.

Hedgemon4 (changes requested):

src/oop/creating_objects_from_classes/hard/q4/MainTest.java
```diff
@@ -0,0 +1,79 @@
...
+        /* Check Method Definitions */
+        assertTrue(testClass.hasMethod("main", mainArgs), String.format(missingMain, testSavingsName, "main"));
+        assertTrue(testClass.hasModifier("main", mainArgs, "public"), incorrectVisibility);
+        assertTrue(testClass.hasModifier("main", mainArgs, "static"), missingStatic);
+        assertTrue(testClass.hasReturnType("main", mainArgs, Void.TYPE), wrongReturnType);
+
+        String incorrectOutput = "Your main method in the %s class does not print the correct output. Make sure that you are printing the balance and calling all required methods.";
+        Clause[] clauses = new Clause[]{
+                new StringLiteral("Balance: "),
+                new DoubleLiteral(initial + firstBonus, initial + firstBonus + 0.00001),
```
Hedgemon4, line 48: Why the + 0.00001 here?

Hedgemon4 (changes requested):

src/oop/creating_objects_from_classes/hard/q4/MainTest.java
SamamthaH: because DoubleLiteral needs two numbers, a lower bound and an upper bound, and the upper bound is exclusive so I can't use the same number, so I added the 0.00001 to add effectively as a delta

PR #625, opened 2023-04-14
arrays-programs_involving_data_sequences-easy-q4

SamamthaH:
## PR Checklist
- [x]  EOF newlines


SamamthaH:
Realized the reason it initially wasn't working was that there were repeat strings, now I've got it mostly working. The if statement at the end is because without it, the 17 and 50 will fail almost every time, but with it it'll pass almost the whole time

src/arrays/programs_involving_data_sequences/easy/q4/MainTest.java
```diff
@@ -0,0 +1,96 @@
...
+        int count = 0;
+
+        for (int x = 0; x < size; x++) {
+            ArrayList<Integer> response = Arrays.stream(numbers[x]).boxed().collect(Collectors.toCollection(ArrayList::new));
+            RandomInteger randomInteger = new RandomInteger(0, size);
+            if (randomInteger.validateRandom(response))
+                count++;
+        }
+
+        String s = "Your musicalChairs method does not randomly shuffle the names correctly.";
```
Hedgemon4, line 88:
```suggestion
        String s = "Your musicalChairs method does not randomly shuffle the names in the array.";
```

Hedgemon4 (approved):

PR #626, opened 2023-04-26
oop-user_defined_classes-medium-q4

KaiKhademi:
## PR Checklist
- [x]  EOF newlines


src/oop/user_defined_classes/medium/q4/QuestionText.md
```diff
@@ -0,0 +1,19 @@
...
+that initializes them.
+
+
+
+## Water
+
+The `Water` class you are given has the following items:
+
+- Attributes:
+  - A `double` called **volume**
```
Hedgemon4, line 16: For this one, we should also add a check to the constructor to see if the volume is not negative.

Hedgemon4 (changes requested):

src/oop/user_defined_classes/medium/q4/QuestionText.md
```diff
@@ -0,0 +1,21 @@
...
+The `Water` class you are given has the following items:
+
+- Attributes:
+    - A `double` called **volume** that is not accessible from another class.
+    - A `String` called **quality** that is not accessible from another class.
+
+You need to add the following items:
+
+- Constructor:
+    - Create a constructor that initializes these attributes in that order. Also, if the input
```
Hedgemon4, line 20: I would specify the order again here and you also need to provide what the accessibility modifier should be

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #630, opened 2023-05-03
oop-creating_objects_from_classes-hard-q1-modifications

Hedgemon4:
## PR Checklist
- [ ]  EOF newlines


src/oop/creating_objects_from_classes/hard/q1/MainTest.java
```diff
@@ -1,59 +1,134 @@
...
+
+class Dog {
+    private String name;
+    private int age;
+    private double weight;
+    private static Dog dog;
+    private static boolean constructorCalled = false;
+    private static String constructorName;
+    private static double constructorWeight;
+    private static int constructorAge;
```
te-sa, line 80: Make those final and leave comment about purpose

te-sa (changes requested):

src/oop/creating_objects_from_classes/hard/q1/MainTest.java
Hedgemon4: You can't do this because these values get changed in the course of the test, so they can't be final

PR #631, opened 2023-05-03
arrays-programs_involving_multidimensional_data-hard-q4

KaiKhademi:
## PR Checklist
- [x]  EOF newlines

### DIFF
--- a/src/arrays/programs_involving_multidimensional_data/hard/q4/MainTest.java
+++ b/src/arrays/programs_involving_multidimensional_data/hard/q4/MainTest.java
@@ -1,0 +1,76 @@
+package arrays.programs_involving_multidimensional_data.hard.q4;
+
+import global.BaseRandomTest;
+import global.MethodTest;
+import global.variables.Clause;
+import global.variables.clauses.NewLine;
+import global.variables.clauses.RandomBoolean;
+import global.variables.clauses.RandomInteger;
+import global.variables.clauses.StringLiteral;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.ArrayList;
+import java.util.stream.Stream;
+
+import static junit.framework.TestCase.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class MainTest extends BaseRandomTest {
+    // Java
+
+    public static final int n = 3;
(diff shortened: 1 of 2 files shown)


src/arrays/programs_involving_multidimensional_data/hard/q4/TicTac.java
```diff
@@ -0,0 +1,42 @@
+package arrays.programs_involving_multidimensional_data.hard.q4;
+
+/*
+In the game of Tic-Tac-Toe is played on a board of size 3 by 3 between two players. Each player has a shape that can be
+used (either a circle 'O' or a cross 'X'), and the player has to get three of the same shapes in a row, column, or a
+diagonal to win. Write a Java program that uses a 2D character array to represent the board and randomly assigns one of
+the shapes to each square. For that, write a method called generateTicTac that
```
Hedgemon4, line 7: You don't ever specify what the passed integer is used for in the method.

Hedgemon4: You need to add something along the lines of "Generate an n by n Tic-Tac-Toe board ... "

src/arrays/programs_involving_multidimensional_data/hard/q4/MainTest.java
```diff
@@ -0,0 +1,76 @@
...
+        return c;
+    }
+
+    public void runMain() {
+        TicTac.main(new String[0]);
+    }
+
+    static Stream<Arguments> inputProvider() {
+        return Stream.of(
+                Arguments.of(40),
```
Hedgemon4, line 45: You should test a some odd numbers, and one that is not divisible by 10

src/arrays/programs_involving_multidimensional_data/hard/q4/MainTest.java
```diff
@@ -0,0 +1,76 @@
...
+            for(int j = 0; j < size; j ++) {
+                if(ans[i][j] == 'X')
+                    arr.add(1);
+                else if(ans[i][j] == 'O')
+                    arr.add(0);
+                else
+                    fail("Your generateTicTac method does not return a random Tic Tac board correctly.");
+            }
+        }
+        RandomInteger rand = new RandomInteger(0, 2);
```
Hedgemon4, line 72: You should use randomBoolean here instead

Hedgemon4 (changes requested):

PR #640, opened 2023-05-09
oop-special_class_method-medium-q7

KaiKhademi:
## PR Checklist
- [x]  EOF newlines


SamamthaH (changes requested):

src/oop/special_class_method/medium/q7/MainTest.java
```diff
@@ -0,0 +1,205 @@
...
+        assertTrue(testClass.hasModifier(classes, "public"), modifiedClassMessage);
+    }
+
+
+    private static Stream<Integer> intInputProvider() {
+        return Stream.of(
+                0, 23, 553233333, 111111, 3554
+        );
+    }
+
```
SamamthaH, line 52: I feel like we should be testing more than one case for the get methods

src/oop/special_class_method/medium/q7/MainTest.java
```diff
@@ -0,0 +1,205 @@
...
+        Object[][] arguments1 = {
+                {value1, int.class}
+        };
+        Object[][] arguments2 = {
+                {value2, int.class}
+        };
+        double value = (value1 * 1 + value2 * 5) / 100.0;
+        String incorrectToStringMessage = "Your toString method does not return the correct String.";
+        Object testInstance = testClass.createInstance(null);
+        testClass.callMethod(setAttributeMethodName1, arguments1, testInstance);
```
SamamthaH, line 141: switch the call methods to setField that way if the set method is incorrect it doesn't interrupt this test

src/oop/special_class_method/medium/q7/MainTest.java
```diff
@@ -0,0 +1,205 @@
...
+        testClass.callMethod(setAttributeMethodName2, arguments2, testInstance);
+        String expectedOutput = toStringMessage + value;
+        Object toStringOutput = testClass.callMethod("toString", testInstance);
+        _assertEquals(expectedOutput, toStringOutput, incorrectToStringMessage);
+    }
+
```
SamamthaH, line 147: I am unsure why you kept the get set methods and made the methods work together since methods work together should contain all the same tests as get set methods making them redundant.

src/oop/special_class_method/medium/q7/MainTest.java
```diff
@@ -0,0 +1,205 @@
...
+    @ParameterizedTest
+    @MethodSource("intInputProvider2")
+    public void methodsWorkingTogether(int value1, int value2) throws Throwable {
+        Object testInstance = testClass.createInstance();
+        Object[][] setMethodArguments = {
+                {value1, int.class}
+        };
+        Object setMethodOutput = testClass.callMethod(setAttributeMethodName1, setMethodArguments, testInstance);
+        _assertEquals(value1, testClass.getFieldValue(testInstance, attributeName1),
```
SamamthaH, line 184: this should be switched to using the get method instead of getFieldValue

SamamthaH: the error message will also need to have the placement of set method and get method reversed

src/oop/special_class_method/medium/q7/MainTest.java
```diff
@@ -0,0 +1,205 @@
...
+                "Your " + setAttributeMethodName1 + " method does not correctly update the value of " + attributeName1 + " after using the " + getAttributeMethodName1 + " method.");
+        assertNull(setMethodOutput, String.join(" ", "Your", setAttributeMethodName1, "method should not return any output"));
+        setMethodArguments = new Object[][]{
+                {value2, int.class}
+        };
+        setMethodOutput = testClass.callMethod(setAttributeMethodName2, setMethodArguments, testInstance);
```
SamamthaH, line 190: same for this

src/oop/special_class_method/medium/q7/MainTest.java
```diff
@@ -0,0 +1,205 @@
...
+        _assertEquals(value2, testClass.getFieldValue(testInstance, attributeName2),
+                "Your " + setAttributeMethodName2 + " method does not correctly update the value of " + attributeName2 + " after using the " + getAttributeMethodName2 + " method.");
+        assertNull(setMethodOutput, String.join(" ", "Your", setAttributeMethodName2, "method should not return any output"));
```
SamamthaH, line 193: no need for assertNull since the set method test will have checked that

src/oop/special_class_method/medium/q7/MainTest.java
```diff
@@ -0,0 +1,205 @@
...
+        Object getMethodOutput = testClass.callMethod(getAttributeMethodName1, testInstance);
+        _assertEquals(value1, getMethodOutput,
+                "Your " + getAttributeMethodName1 + " method does not return the value of the " + attributeName1 + " attribute.");
```
SamamthaH, line 196: this will have already been tested once getFieldValue is swapped

SamamthaH (changes requested):

src/oop/special_class_method/medium/q7/MainTest.java
KaiKhademi: Since the constructor only assigns them to 0, we can't test them without using the setters, which will also test the validity of setters and we don't want that. We already have more than one test in worksTogether for the getter.

KaiKhademi: Never mind I can just use the setField.

SamamthaH (dismissed):

Hedgemon4 (changes requested):
See feedback on other questions in this category

Hedgemon4 (approved):

PR #644, opened 2023-05-11
oop-user_defined_classes-hard-q2

SamamthaH:
## PR Checklist
- [x]  EOF newlines


SamamthaH:
is this section supposed to have comments like // Java at the top? Because the example didn't

src/oop/user_defined_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,116 @@
...
+    public void setUp() {
+        String testClassString = "oop.user_defined_classes.hard.q2." + testClassName;
+        classInstance = new ObjectTest(testClassString);
+    }
+
+    @Test
+    public void dishWasherClassHasCorrectFields() {
+        String missingFieldMessage = "Your " + testClassName + " is missing a required field";
+        String incorrectModifierMessage = "One of your fields in the " + testClassName +
+                " class does not have the correct visibility modifier.";
```
KaiKhademi, line 37: Change the "fields" to "attributes" going forward.

src/oop/user_defined_classes/hard/q2/QuestionText.md
```diff
@@ -0,0 +1,30 @@
...
+## Notes
+
+- Vary both of the attributes for this question and the class name
+- Also vary the method name
+
+## Question Text
+
+Dishwashers are convenient, but it can be easy to lose track of how much clean cutlery you have. Write a Java class 
+called `DishWasher` to track the contents of the dishwasher. See the description below for the requirements.
+
```
KaiKhademi, line 12: Same here "fields" should be "attributes"

src/oop/user_defined_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,116 @@
...
+                Arguments.of(378, 32891, 378, 32891, 0, 0),
+                Arguments.of(0, 0, 0, 0, 0, 0),
+                Arguments.of(5, 10, 0, 1, 5, 9)
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("methodInputProvider")
+    public void correctLoadDishWasherMethod(int attribute1, int attribute2, int remove1, int remove2, int remainder1, int remainder2) throws Throwable{
+        Object[][] arguments = {
```
KaiKhademi, line 94: I don't think you actually need the remainders here as parameters you can check for attribute - remove

src/oop/user_defined_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,116 @@
...
+        assertTrue(classInstance.hasMethod(methodName, new Class[]{int.class, int.class}),
+                "Your " + testClassName + " class is missing the method " + methodName + ".");
+        classInstance.callMethod(
+                methodName, new Object[][]{{remove1, int.class}, {remove2, int.class}}, null, instance, new Clause[]{
+                        new IntegerLiteral(remainder1),
+                        new StringLiteral(" " + intAttributeName1 + " are still clean"),
+                        new NewLine(),
+                        new IntegerLiteral(remainder2),
+                        new StringLiteral(" " + intAttributeName2 + " are still clean"),
+                        new NewLine()
```
KaiKhademi, line 108: I think its better to add optional space at the end of the outputs

KaiKhademi (changes requested):

src/oop/user_defined_classes/hard/q2/MainTest.java
SamamthaH: Yes, but it effectively achieves the same end and I prefer to minimize computer calculation when possible. Removing the remainder attributes and swapping them out for a calculation doesn't improve anything I don't believe.

src/oop/user_defined_classes/hard/q2/MainTest.java
SamamthaH: I checked and it looks like the clause testing adds an optional newLine at the end anyways so I just got rid of my new line and let that invisible one take its place

src/oop/user_defined_classes/hard/q2/QuestionText.md
```diff
@@ -0,0 +1,30 @@
...
+### DishWasher Class:
+
```
KaiKhademi, line 14: For all of these also add what visibility modifiers they need for example:
An int called forks which is not visible from other classes.
meaning that it should be private.

src/oop/user_defined_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,115 @@
...
+        String testClassString = "oop.user_defined_classes.hard.q2." + testClassName;
+        classInstance = new ObjectTest(testClassString);
+    }
+
+    @Test
+    public void dishWasherClassHasCorrectAttributes() {
+        String missingAttributeMessage = "Your " + testClassName + " is missing a required attribute";
+        String incorrectModifierMessage = "One of your attributes in the " + testClassName +
+                " class does not have the correct visibility modifier.";
+        assertTrue(classInstance.hasField(intAttributeName1, int.class),
```
KaiKhademi, line 38: When testing for the attributes we should mention the attributes that are missing in the error message same for the modifies.
Something like:
"Your " + className + " class is missing the " + attribute + " attribute."
"Your " + className + " class " + attribute + " attributes does not have the correct modifier."

src/oop/user_defined_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,115 @@
...
+                incorrectModifierMessage);
+    }
+
+    @Test
+    public void dishWasherClassHasRequiredConstructor() {
+        String missingConstructorMessage = "Your " + testClassName + " class is missing a required constructor.";
+        Class<?>[] arguments= {
+                int.class, int.class
+        };
+        assertTrue(classInstance.hasConstructor(arguments), missingConstructorMessage);
```
KaiKhademi, line 54: Here you should also check that the constructor has the correct modifier

src/oop/user_defined_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @ParameterizedTest
+    @MethodSource("methodInputProvider")
+    public void correctLoadDishWasherMethod(int attribute1, int attribute2, int remove1, int remove2, int remainder1, int remainder2) throws Throwable{
+        Object[][] arguments = {
+                {attribute1, int.class},
+                {attribute2, int.class}
+        };
+        Object instance = classInstance.createInstance(arguments);
+        assertTrue(classInstance.hasMethod(methodName, new Class[]{int.class, int.class}),
```
KaiKhademi, line 99: I think here you can also check the return type and modifiers.

KaiKhademi (changes requested):

src/oop/user_defined_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @Test
+    public void dishWasherClassHasRequiredConstructor() {
+        Class<?>[] arguments= {
+                int.class, int.class
+        };
+        assertTrue(classInstance.hasConstructor(arguments),
+                String.format("Your %s class is missing a required constructor.", testClassName));
+        assertTrue(classInstance.hasConstructor(arguments, new String[]{"public"}),
```
KaiKhademi, line 55: I think it's better if you use hasModifier() here because if the arguments in the constructor are wrong but it has the correct modifier this test would fail and say that the modifier is incorrect.

KaiKhademi (changes requested):

src/oop/user_defined_classes/hard/q2/DishWasher.java
```diff
@@ -0,0 +1,18 @@
+package oop.user_defined_classes.hard.q2;
+
+public class DishWasher {
+    private int forks;
+    private int knives;
+    public DishWasher(int a, int b)
```
Hedgemon4, line 6: Please format this file

Hedgemon4 (changes requested):

KaiKhademi (approved):

src/oop/user_defined_classes/hard/q2/QuestionText.md
```diff
@@ -0,0 +1,31 @@
...
+    - An `int` called **forks** which is not visible from other classes
+    - An `int` called **knives** which is not visible from other classes
+
+- Constructors:
+    - A constructor which accepts two ints and sets the **forks** and **knives** attributes. The first int should
+      set **forks** attribute and the second the **knives** attribute.
+    - This constructor should be visible to other classes
+
+- Methods:
+    - A method called `loadDishWasher` which another class can give the number of dirty forks and dirty knives to be
```
Hedgemon4, line 27:
```suggestion
    - A method called `loadDishWasher` where another class can give the number of dirty forks and dirty knives to be
```

src/oop/user_defined_classes/hard/q2/QuestionText.md
```diff
@@ -0,0 +1,31 @@
...
+      subtracted from the total number of **forks** and **knives**. The first int should set **forks** attribute and the
```
Hedgemon4, line 28: I would use something other than "set" in this context, I feel like it could be confusing

src/oop/user_defined_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,120 @@
...
+    public void setUp() {
+        String testClassString = "oop.user_defined_classes.hard.q2." + testClassName;
+        classInstance = new ObjectTest(testClassString);
+    }
+
+    @Test
+    public void dishWasherClassHasCorrectAttributes() {
+        String missingAttributeMessage =
+                "The attribute %s could not be found in your class. Please make sure you have added it and that it is spelled correctly.";
+        String wrongTypeMessage = "The attribute %s does not have the correct type.";
```
Hedgemon4, line 37: Can we say "The x attribute ..." instead of  "The attribute x ..." ? I feel like it sounds more grammatically correct.

src/oop/user_defined_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,120 @@
...
+    @ParameterizedTest
+    @MethodSource("constructorInputProvider")
+    public void dishWasherConstructorInitializesAttributesCorrectly(int attribute1, int attribute2) throws Throwable {
+        Object[][] arguments = {
+                {attribute1, int.class},
+                {attribute2, int.class}
+        };
+        Object instance = classInstance.createInstance(arguments);
+        String incorrectAttributeInstantiationMessage =
+                "Your " + testClassName + " constructor does not correctly initialize the object.";
```
Hedgemon4, line 79: I would instead say which attribute it did not initialize correctly.

src/oop/user_defined_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,120 @@
...
+    @ParameterizedTest
+    @MethodSource("methodInputProvider")
+    public void correctLoadDishWasherMethod(int attribute1, int attribute2, int remove1, int remove2, int remainder1, int remainder2) throws Throwable {
+        Object[][] arguments = {
+                {attribute1, int.class},
+                {attribute2, int.class}
+        };
+        Object instance = classInstance.createInstance(arguments);
+        assertTrue(classInstance.hasMethod(methodName, new Class[]{int.class, int.class}, Void.TYPE, new String[]{"public"}),
+                "Your " + testClassName + " class is missing the method " + methodName + " or it does not have the correct return type or visibility modifier.");
```
Hedgemon4, line 105: This checks too many things in one assertion. Check the return type and visibility modifiers separately. This should also make the error messages more cohesive.

src/oop/user_defined_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,120 @@
...
+        classInstance.callMethod(
+                methodName, new Object[][]{{remove1, int.class}, {remove2, int.class}}, null, instance, new Clause[]{
+                        new IntegerLiteral(remainder1),
+                        new StringLiteral(" " + intAttributeName1 + " are still clean"),
+                        new NewLine(),
+                        new IntegerLiteral(remainder2),
+                        new StringLiteral(" " + intAttributeName2 + " are still clean")
+                }, "Your " + methodName + " method does not print the correct messages."
```
Hedgemon4, line 113:
```suggestion
                }, "Your " + methodName + " method does not print the correct output."
```

Hedgemon4 (changes requested):

src/oop/user_defined_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,124 @@
...
+                {attribute1, int.class},
+                {attribute2, int.class}
+        };
+        Object instance = classInstance.createInstance(arguments);
+        assertTrue(classInstance.hasMethod(methodName, new Class[]{int.class, int.class}),
+                "Your " + testClassName + " class is missing the method " + methodName + ". Check that is spelled correctly and has the correct parameters.");
+        assertTrue(classInstance.hasMethod(methodName, new Class[]{int.class, int.class}, Void.TYPE),
+                "Your " + testClassName + " " + methodName + " method does not have the correct return type.");
+        assertTrue(classInstance.hasMethod(methodName, new Class[]{int.class, int.class}, Void.TYPE, new String[]{"public"}),
+                "Your " + testClassName + " " + methodName + " method does not have the correct return type visibility modifier.");
```
Hedgemon4, line 109:
```suggestion
                "Your " + testClassName + " " + methodName + " method does not have the correct visibility modifier.");
```

src/oop/user_defined_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,124 @@
...
+        classInstance.callMethod(
+                methodName, new Object[][]{{remove1, int.class}, {remove2, int.class}}, null, instance, new Clause[]{
+                        new IntegerLiteral(remainder1),
```
Hedgemon4, line 112: Maybe don't test the exact int here, otherwise, if they are not updating the value properly, I think it will just say they are not printing the correct output.

Hedgemon4: You can test that they are printing the correct number after.

src/oop/user_defined_classes/hard/q2/MainTest.java
```diff
@@ -0,0 +1,124 @@
...
+                        new StringLiteral(" " + intAttributeName1 + " are still clean"),
+                        new NewLine(),
+                        new IntegerLiteral(remainder2),
+                        new StringLiteral(" " + intAttributeName2 + " are still clean")
+                }, "Your " + methodName + " method does not print the correct output."
+        );
+        _assertEquals(remainder1, classInstance.getFieldValue(instance, intAttributeName1),
```
Hedgemon4, line 119: Like this check should come before the print check for the right values.

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #676, opened 2023-06-03
oop-object_interactions-hard-q4

SamamthaH:
## PR Checklist
- [x]  EOF newlines

### DIFF
--- a/src/oop/object_interactions/hard/q4/MainTest.java
+++ b/src/oop/object_interactions/hard/q4/MainTest.java
@@ -33,38 +33,6 @@ public class MainTest {
         classInstance = new ObjectTest(objectString);
         String testString = "oop.object_interactions.hard.q4." + testClassName;
         testInstance = new ObjectTest(testString);
-
-//        assertTrue(testClass.hasModifier("public"),
-//                "You have changed the visibility modifier on the " + className + " class. Please revert it to the original state.");
-//        assertTrue(testClass.hasField(myYearAttributeName, int.class, new String[]{"private"}),
-//                String.format("You have modified the provided %s attribute. Please revert it to the original state.", myYearAttributeName));
-//        assertTrue(testClass.hasField(myMonthAttributeName, int.class, new String[]{"private"}),
-//                String.format("You have modified the provided %s attribute. Please revert it to the original state.", myMonthAttributeName));
-//        assertTrue(testClass.hasField(myDayAttributeName, int.class, new String[]{"private"}),
-//                String.format("You have modified the provided %s attribute. Please revert it to the original state.", myDayAttributeName));
-//        assertTrue(testClass.hasField(myStoryAttributeName, String.class, new String[]{"private"}),
-//                String.format("You have modified the provided %s attribute. Please revert it to the original state.", myStoryAttributeName));
-//        assertTrue(testClass.hasConstructor(null, new String[]{"public"}),
-//                "You have modified the provided constructor on the " + className + " class. Please revert it to the original state.");
-//        Object classInstance = testClass.createInstance();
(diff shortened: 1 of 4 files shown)


src/oop/object_interactions/hard/q4/QuestionText.md
```diff
@@ -0,0 +1,44 @@
...
+- Java question
+
+## Question Text
+
+Some data should be encrypted, however to do so you need a key to base the encryption off of. Design part of a `CryptBox`
+class using the provided class components, and create a `crossEncrypt` method that takes another `CryptBox` object and 
+uses that object's key to encrypt the other object's data.
+
+### CryptBox Class:
+
```
KaiKhademi, line 12: This is a hard level assignment so I guess they can write the whole object. If you want them to write it partially their tests should be in setup(). But here its better to just ask them to write it all.

src/oop/object_interactions/hard/q4/QuestionText.md
```diff
@@ -0,0 +1,44 @@
...
+    - Both of these attributes should not be accessible from another class
+
+- Constructor:
+    - A constructor visible from another class that accepts a `String` and an `int` parameter, and initializes the 
+      **lockedData** attribute with the `String` and the **cryptKey** attribute with the `int`
+    - This constructor should be visible from another class
+
+- Methods
+    - A `getCryptKey` method which returns the **cryptKey** attribute of the `CryptBox`
+
```
KaiKhademi, line 27: Maybe change the visibility modifier wording of the items that are already given to "is visible" or "is not accessible" rather the "should be visible" or "should be accessible"

src/oop/object_interactions/hard/q4/MainTest.java
```diff
@@ -0,0 +1,137 @@
...
+    @ParameterizedTest
+    @MethodSource("constructorInputProvider")
+    public void correctGetCryptKeyMethod(String data, int key) throws Throwable {
+        Object[][] constructorArguments = {
+                {data, String.class},
+                {key, int.class}
+        };
+        Object instance = classInstance.createInstance(constructorArguments);
+        String incorrectGetMethods = "Your %s method does not return the value of the %s attribute.";
+        assertTrue(classInstance.hasMethod(getIntAttribute1, null, int.class, new String[]{"public"}),
```
KaiKhademi, line 86: I believe this is redundant as the callMethod and assertion check all these

KaiKhademi (changes requested):

src/oop/object_interactions/hard/q4/QuestionText.md
SamamthaH: due to meeting, switching this to parsons instead

src/oop/object_interactions/hard/q4/MainTest.java
```diff
@@ -0,0 +1,142 @@
...
+    @ParameterizedTest
+    @MethodSource("constructorInputProvider")
+    public void cryptBoxClassHasCorrectConstructor(String data, int key) throws Throwable {
+        Object[][] arguments = {
+                {data, String.class},
+                {key, int.class}
+        };
+        assertTrue(classInstance.hasConstructor(new Class[]{String.class, int.class}),
+                String.format("Your %s class is missing a required constructor.", objectClassName));
+        assertTrue(classInstance.hasConstructor(new Class[]{String.class, int.class}, new String[]{"public"}),
```
KaiKhademi, line 74: You don't need these two here you are already testing them

KaiKhademi (changes requested):

Hedgemon4 (approved):

PR #682, opened 2023-06-10
oop-static-modifier-medium-q3

SamamthaH:
## PR Checklist
- [x]  EOF newlines

### DIFF
--- a/src/oop/static_modifier/medium/q3/MainTest.java
+++ b/src/oop/static_modifier/medium/q3/MainTest.java
@@ -2,7 +2,6 @@ package oop.static_modifier.medium.q3;
 
 import global.ObjectTest;
 import global.variables.Clause;
-import global.variables.clauses.IntegerLiteral;
 import global.variables.clauses.NewLine;
 import global.variables.clauses.StringLiteral;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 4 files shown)


src/oop/static_modifier/medium/q3/MainTest.java
```diff
@@ -0,0 +1,230 @@
...
+    @BeforeEach
+    public void setup() {
+        String classPath = "oop.static_modifier.medium.q3." + objectName;
+        classInstance = new ObjectTest(classPath);
+        String testPath = "oop.static_modifier.medium.q3." + testName;
+        testInstance = new ObjectTest(testPath);
+    }
+
+    @Test
+    public void horseClassHasRequiredAttributes() {
```
Hedgemon4, line 43: I think you will also need to test here that the static attribute is set to "Neigh" and you will probably have to order the tests as well.

src/oop/static_modifier/medium/q3/MainTest.java
```diff
@@ -0,0 +1,230 @@
...
+                "Your " + objectName + " class toString method does not have the correct return type.");
+        assertTrue(classInstance.hasMethod("toString", new Class[]{}, String.class, new String[]{"public"}),
+                "Your " + objectName + " class toString method does not have the correct visibility modifier.");
+        String expectedOutput = name + " calls out " + startNoise;
+        Object output = classInstance.callMethod("toString", new String[]{"public"}, instance);
+        String incorrectToString = String.format("Your toString method for the %s class does return the correct String.", objectName);
+        _assertEquals(expectedOutput, output, incorrectToString);
+    }
+
+    private static Stream<Arguments> methodInputProvider() {
```
Hedgemon4, line 107: Should we also test if there are no vowels in the word? Or does this result in undefined behaviour?

Hedgemon4 (changes requested):

Hedgemon4 (approved):
Fixed issues

PR #708, opened 2023-08-08
methods-modular_programs-medium-q3

KaiKhademi:
## PR Checklist
- [x]  EOF newlines


src/methods/modular_programs/medium/q3/AlternateIt.java
```diff
@@ -0,0 +1,29 @@
+package methods.modular_programs.medium.q3;
+
+/**
+ * *Write a Java program with a method called sumCalc to ask the user for a
```
Hedgemon4, line 4: Please specify what parts of the question are to be done in the main method and what parts of the question are to be done in the sumCalc method. I think it is unclear currently.

Hedgemon4: You also do not specify the parameters and the return type of the method.

src/methods/modular_programs/medium/q3/MainTest.java
```diff
@@ -0,0 +1,68 @@
...
+                Arguments.of(992, -496));
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputUpperBoundProvider")
+    public void printsCorrectOutput(int upperBound, int expectedSum) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not print the correct message of the series sum.";
+        runWithInput(upperBound + "", new Clause[][]{{
+                new StringLiteral("The sum of the series is: "),
+                new IntegerLiteral(expectedSum),
```
Hedgemon4, line 52: Please check this outside of the run with input. You can leave the expected value blank and catch any number, then you can check that it prints the correct value.

Hedgemon4: I think we can then provide a better error message to the students, one more specifically for the print format and one if the number is wrong, as that is likely to be caused by their method being incorrect.

src/methods/modular_programs/medium/q3/MainTest.java
```diff
@@ -0,0 +1,68 @@
...
+        }});
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputUpperBoundProvider")
+    public void correctSumCalcMethod(int n, int ans) throws Throwable {
+        Clause[] methodSentence = new Clause[]{
```
Hedgemon4, line 59: You can also do a similar thing here where you can set an error message for if the method prints the correct output and then check the printed number after as well. Take a look at the `MethodTest` docs if you do not know how to do this and reach out to me if that does not help.

Hedgemon4 (changes requested):

Hedgemon4 (approved):

PR #761, opened 2025-04-13
Conditionals nested statements hard q11

AdaraPutri:
## PR Checklist
- [ ]  EOF newlines


src/conditionals/nested_statements/q11/FruitIdentifier.java
```diff
@@ -0,0 +1,44 @@
+package conditionals.nested_statements.q11;
+import java.util.Scanner;
+
+/*
+ Rearrange the following Java code snippets to correctly implement a nested switch statement that determines 
+ an item based on two input Strings, color and taste. The outer switch should handle color, and the inner 
+ switch should handle taste within certain cases. The output should be a single string indicating the item.
```
SamamthaH, line 7:
```suggestion
 switch should handle taste within certain cases. The output should be a single string indicating the item matching those descriptors.
```

src/conditionals/nested_statements/q11/FruitIdentifier.java
```diff
@@ -0,0 +1,44 @@
+package conditionals.nested_statements.q11;
+import java.util.Scanner;
+
+/*
+ Rearrange the following Java code snippets to correctly implement a nested switch statement that determines 
+ an item based on two input Strings, color and taste. The outer switch should handle color, and the inner
```
SamamthaH, line 6:
```suggestion
 an item based on two input Strings giving it's color and taste. The outer switch should handle color, and the inner 
```

src/conditionals/nested_statements/q11/FruitIdentifier.java
```diff
@@ -0,0 +1,44 @@
...
+                       break;
+               }
+               break;
+           case "Yellow":
+               switch (taste) {
+                   case "Sweet":
+                       System.out.print("Banana");
+                       break;
+                   default:
+                       break;
```
SamamthaH, line 39: add another "not found" here

src/conditionals/nested_statements/q11/FruitIdentifier.java
```diff
@@ -0,0 +1,44 @@
...
+ Sample output:
+ Chili
+ */
+
+
+public class FruitIdentifier {
+   public static void main(String[] args) {
+       String color = "Red";
+       String taste = "Spicy";
+
```
SamamthaH, line 18: I think this does need to be an input test. Otherwise students can just work out the problem and put in the class and method initialization then the Chili print statement without arranging the switch at all

SamamthaH (changes requested):

src/conditionals/nested_statements/q11/FruitIdentifier.java
AdaraPutri: added!

src/conditionals/nested_statements/q11/FruitIdentifier.java
AdaraPutri: I see what you mean, that's a good point! I've changed the question file and MainTest file to accommodate the input test conversion.

src/conditionals/nested_statements/hard/q11/FruitIdentifier.java
```diff
@@ -0,0 +1,56 @@
...
+
+
+public class FruitIdentifier {
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        System.out.println("Enter color: ");
+        String color = scanner.nextLine();
+        System.out.println("Enter taste: ");
+        String taste = scanner.nextLine();
+        scanner.close();
```
SamamthaH, line 28:
```suggestion
```

SamamthaH (approved):

PR #762, opened 2025-04-13
Conditionals nested statements hard q12

AdaraPutri:
## PR Checklist
- [ ]  EOF newlines

### DIFF
--- a/src/conditionals/nested_statements/hard/q12/MainTest.java
+++ b/src/conditionals/nested_statements/hard/q12/MainTest.java
@@ -4,7 +4,6 @@ import global.BaseTest;
 import global.exceptions.InvalidClauseException;
 import global.tools.TestOption;
 import global.variables.Clause;
-import global.variables.clauses.NewLine;
 import global.variables.clauses.PlaceHolder;
 import global.variables.clauses.StringLiteral;
 import org.junit.jupiter.params.ParameterizedTest;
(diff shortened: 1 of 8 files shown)


src/conditionals/nested_statements/q12/ScoreAnalyzer.java
```diff
@@ -0,0 +1,54 @@
+package conditionals.nested_statements.q12;
+
```
SamamthaH, line 2: Add question text

src/conditionals/nested_statements/q12/ScoreAnalyzer.java
```diff
@@ -0,0 +1,54 @@
...
+public class ScoreAnalyzer {
+    // Parsons
```
SamamthaH, line 4: This comment should be in the MainTest file not the question file

src/conditionals/nested_statements/q12/ScoreAnalyzer.java
```diff
@@ -0,0 +1,54 @@
...
+    public static void main(String[] args) {
+        int score = 87;
+
+        switch (score % 2) {
+            case 0:
+                System.out.println("Even Score");
+
+                switch (score / 10) {
```
SamamthaH, line 12: why are these other switch statements nested? nothing about them changes between even and odd

src/conditionals/nested_statements/q12/MainTest.java
```diff
@@ -0,0 +1,33 @@
+package conditionals.nested_statements.q12;
```
SamamthaH, line 1: there should also be a package between nested_statements and q12 for difficult. I didn't spot this in the other pr but it applies there as well

src/conditionals/nested_statements/q12/MainTest.java
```diff
@@ -0,0 +1,33 @@
...
+        };
+    }
+
+    public void runMain() {
+        ScoreAnalyzer.main(new String[0]);
+    }
+
+    @Test
+    public void ScoreAnalyzeTest() {
+        String expected = line1 + System.lineSeparator() + line2 + System.lineSeparator();
```
SamamthaH, line 30: this is already tested by the testSentence

SamamthaH (changes requested):
This question also has the issue of because it's not an input test students can just solve the question and print the two lines they know they need to and ignore the switch

src/conditionals/nested_statements/q12/ScoreAnalyzer.java
AdaraPutri: I see your point, the code block is definitely redundant. But this category specifically required the use of a nested switch statement. I've updated the printed outputs so they vary based on whether the number is odd or even, but let me know if you have any suggestions for a more meaningful or relevant nested switch statement idea!

src/conditionals/nested_statements/hard/q12/ScoreAnalyzer.java
```diff
@@ -0,0 +1,60 @@
+package conditionals.nested_statements.hard.q12;
+import java.util.Scanner;
+/*
+ * Rearrange the following Java code snippets in the correct order to correctly implement a nested switch
+ * statement that determines a grade category based on a score.
```
SamamthaH, line 5: they need to know which print statements belong to which conditions

src/conditionals/nested_statements/q12/ScoreAnalyzer.java
SamamthaH: You could have it be something like "divide students into groups based on their grade" so odds would be group 1 + letter grade and evens would be group 2 + letter grade and you just keep it all on one print line instead of splitting up the print lines

SamamthaH (changes requested):

src/conditionals/nested_statements/hard/q12/MainTest.java
```diff
@@ -0,0 +1,55 @@
...
+                Arguments.of("61", "Group 1 - D"),
+                Arguments.of("59", "Group 1 - F"),
+                Arguments.of("50", "Group 2 - F")
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void testScoreCategory(String grade, String expectedOutput) throws InvalidClauseException {
+        TestOption.incorrectStructureErrorMessage = "Your program does not correctly output the group and grade.";
```
SamamthaH, line 49: error message no longer makes sense

SamamthaH (approved):

PR #763, opened 2025-04-13
Conditionals simple programs hard q9

AdaraPutri:
## PR Checklist
- [ ]  EOF newlines

### DIFF
--- a/src/conditionals/simple_programs_with_decision_points/hard/q9/GradeToLetter.java
+++ b/src/conditionals/simple_programs_with_decision_points/hard/q9/GradeToLetter.java
@@ -3,13 +3,7 @@ package conditionals.simple_programs_with_decision_points.hard.q9;
 import java.util.Scanner;
 
 /*
- Write a Java program that prompts the user to enter a numerical grade and outputs the corresponding letter grade.
- 
- Your program must use a switch statement to determine the letter based on the input grade.
-
- Use the sentence:
-     "Enter the numerical grade: "
- as your prompt to the user.
+ Write a Java program that asks the user to "Enter the numerical grade: " and uses a switch statement to output the corresponding letter grade on its own line. 
 
  Grading scale:
  - 90 and above: A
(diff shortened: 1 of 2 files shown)


src/conditionals/simple_programs_with_decision_points/hard/q9/GradeToLetter.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.Scanner;
+
+/*
+ Suppose there is a Java program that takes <tt>grade</tt> as integer input from the user and prints the letter correspondingly 
+ using the if-else statement below. Translate the following statement into a switch statement block, and use the sentence 
+ “Enter the numerical grade: “ to ask for input.
+
+<code>
+        if (grade >= 90) {
```
SamamthaH, line 11: I don't think we should give them this code block because they could just copy paste the code block in and get full marks with no effort.

src/conditionals/simple_programs_with_decision_points/hard/q9/MainTest.java
```diff
@@ -0,0 +1,61 @@
...
+                Arguments.of("60", "D"),
+                Arguments.of("59", "F"),
+                Arguments.of("30", "F"),
+                Arguments.of("100", "A")
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("inputProvider")
+    void printsCorrectWeather(String inputGrade, String expectedOutput) throws InvalidClauseException {
```
SamamthaH, line 54:
```suggestion
    void printsCorrectLetter(String inputGrade, String expectedOutput) throws InvalidClauseException {
```

SamamthaH (changes requested):

src/conditionals/simple_programs_with_decision_points/hard/q9/GradeToLetter.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Scanner;
+
+/*
+ Write a Java program that prompts the user to enter a numerical grade and outputs the corresponding letter grade.
+ 
+ Your program must use a switch statement to determine the letter based on the input grade.
+
+ Use the sentence:
+     "Enter the numerical grade: "
+ as your prompt to the user.
```
SamamthaH, line 12:
```suggestion
 Write a Java program that asks the user to "Enter the numerical grade: " and uses a switch statement to output the corresponding letter grade on its own line. 
```

SamamthaH (approved):

PR #765, opened 2025-04-14
Programs with simple decision points q14

jae186:
## PR Checklist
- [ ]  EOF newlines

### DIFF
--- a/src/conditionals/programs_with_decision_points/hard/q14/MainTest.java
+++ b/src/conditionals/programs_with_decision_points/hard/q14/MainTest.java
@@ -15,38 +15,38 @@ import org.junit.jupiter.params.provider.MethodSource;
 import java.util.stream.Stream;
 
 public class MainTest extends BaseTest {
-    
+//Parsons
     @Override
     public Clause[] testSentence() {
         TestOption.isInputTest = true;
         TestOption.defaultInput = "apple\nbanana";
         
         return new Clause[]{
-                new StringLiteral("Enter word1 "),
+                new StringLiteral("Enter word1:"),
                 new Optional(new StringLiteral(" ")),
                 new NewLine(),
-                new StringLiteral("Enter word2 "),
+                new StringLiteral("Enter word2:"),
                 new Optional(new StringLiteral(" ")),
                 new NewLine(),
                 new PlaceHolder()
         };
     }
     
-    /**
-     * Provides test cases as (word1, word2, expectedPhrase).<br>
-     * The expected outputs are defined according to:
-     * <ul>
-     *    <li>Combined length < 10: "Short phrase."</li>
-     *    <li>Combined length between 10 and 15 (inclusive): "Medium phrase."</li>
-     *    <li>Combined length > 15: if word1 starts with lowercase 'a', "a-start long phrase." otherwise "Long phrase."</li>
-     * </ul>
-     */
+
     static Stream<Arguments> inputProvider() {
         return Stream.of(
                 // Test: total length 2+3 = 5 (less than 10)
                 Arguments.of("hi", "hey", "Short phrase."),
+                // Test: total length 2+7 = 9 (less than 10)
+                Arguments.of("hi", "heyyyyy", "Short phrase."),
                 // Test: total length 5+5 = 10 (medium)
                 Arguments.of("hello", "world", "Medium phrase."),
+                // Test: total length 5+6 = 11 (medium)
+                Arguments.of("hello", "worldd", "Medium phrase."),
+                // Test: total length 5+9 = 14 (medium)
+                Arguments.of("hello", "worlddddd", "Medium phrase."),
+                // Test: total length 5+11 = 16 (long and word1 does not with lowercase 'a')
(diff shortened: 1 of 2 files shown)


src/conditionals/programs_with_decision_points/hard/q14/checkWord.java
```diff
@@ -0,0 +1,58 @@
+package conditionals.programs_with_decision_points.hard.q14;
+/*
+Write a program that asks the user for two words, word1 and word2.
+Then, check if the combined length of both words:
+Less than 10: Print "Short phrase."
+Between 10 and 15 (inclusive): Print "Medium phrase."
+Greater than 15: If word1 starts with lowercase "a", print "a-start long phrase."
+Otherwise, print "Long phrase."
```
SamamthaH, line 8:
```suggestion
Write a program that asks the user for two words, word1 and word2. Then check if the combined length of both words is less than 10, if so print "Short phrase." If it's between 10 and 15 (inclusive) print "Medium phrase." If it is greater than 15 and word1 starts with lowercase "a", print "a-start long phrase." Otherwise, print "Long phrase."
```

src/conditionals/programs_with_decision_points/hard/q14/checkWord.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+import java.util.Scanner;
+
+public class checkWord {
```
SamamthaH, line 12: class name should have the c capitalized as well

src/conditionals/programs_with_decision_points/hard/q14/checkWord.java
```diff
@@ -0,0 +1,58 @@
...
+
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+
+        System.out.println("Enter word1 ");
```
SamamthaH, line 17: Same as last pr, ad colons for site-wide consistency (also why I've been reformatting the questions)

src/conditionals/programs_with_decision_points/hard/q14/MainTest.java
```diff
@@ -0,0 +1,73 @@
...
+import global.variables.clauses.StringLiteral;
+import global.variables.wrappers.Optional;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
+
+public class MainTest extends BaseTest {
+
```
SamamthaH, line 18: note the question type

src/conditionals/programs_with_decision_points/hard/q14/checkWord.java
SamamthaH: This question doesn't feel set up in a way to prompt students to use a switch statement, the fact that even the solution initially uses an if statement tells me this needs to get reworked a bit.

src/conditionals/programs_with_decision_points/hard/q14/MainTest.java
```diff
@@ -0,0 +1,73 @@
...
+public class MainTest extends BaseTest {
+    
+    @Override
+    public Clause[] testSentence() {
+        TestOption.isInputTest = true;
+        TestOption.defaultInput = "apple\nbanana";
+        
+        return new Clause[]{
+                new StringLiteral("Enter word1 "),
```
SamamthaH, line 25: remove extra space here and in second basic stringliteral because the extra space exists in the optional

src/conditionals/programs_with_decision_points/hard/q14/MainTest.java
```diff
@@ -0,0 +1,73 @@
...
+     * Provides test cases as (word1, word2, expectedPhrase).<br>
+     * The expected outputs are defined according to:
+     * <ul>
+     *    <li>Combined length < 10: "Short phrase."</li>
+     *    <li>Combined length between 10 and 15 (inclusive): "Medium phrase."</li>
+     *    <li>Combined length > 15: if word1 starts with lowercase 'a', "a-start long phrase." otherwise "Long phrase."</li>
+     * </ul>
+     */
+    static Stream<Arguments> inputProvider() {
+        return Stream.of(
```
SamamthaH, line 45: same as last pr, you should also add tests on either side of the boundaries for each threshold

SamamthaH (changes requested):

src/conditionals/programs_with_decision_points/hard/q14/checkWord.java
jae186: That's fair, I forgot to write it but it's a Parsons which I thought would help. Plus giving the chronological directions of first finding the length I thought I would help. Also this is a good way to use if statements and switch statements together which I thought might be nice since it's the hardest switch category.

jae186:
Thanks for the feedback Sam, this one is ready again! @SamamthaH

SamamthaH (approved):

